PR #5, opened 2020-03-25
StartAt logic moved to scheduler.go

Streppel:


Streppel:
@JohnRoesler never mind the commit history I accidentally created this branch from yours

gocron_test.go
```diff
@@ -90,36 +90,15 @@ func TestScheduler_WeekdaysTodayAfter(t *testing.T) {
 	}
 }
 
-func TestScheduleNextRunLoc(t *testing.T) {
```
Streppel: this test didn't make sense anymore since we don't rely on jobs timezone anymore

gocron_test.go
```diff
@@ -90,36 +90,15 @@ func TestScheduler_WeekdaysTodayAfter(t *testing.T) {
...
-	assert.Equal(t, tomorrow.Day(), job.NextScheduledTime().UTC().Day())
-}
-
 func TestScheduleNextRunFromNow(t *testing.T) {
 	sched := NewScheduler(time.UTC)
 	now := time.Now().UTC()
 
-	job, _ := sched.Every(1).Hour().From(NextTick()).Do(task)
-
+	job, _ := sched.Every(1).Hour().StartImmediately().Do(task)
```
Streppel: I've also changed this nomenclature to start a job upon scheduler start. it can be changed if you have a better name in mind

JohnRoesler: No, I like that better

scheduler.go
```diff
@@ -280,10 +279,17 @@ func (j *Job) GetAt() string {
...
-func (s *Scheduler) From(t *time.Time) *Scheduler {
-	currentJob := s.getCurrentJob()
-	currentJob.nextRun = *t
+// StartAt schedules the next run of the job
+func (s *Scheduler) StartAt(t time.Time) *Scheduler {
+	s.getCurrentJob().nextRun = t
+	return s
+}
+
+// NextTick returns a pointer to a time that will run at the next tick
```
JohnRoesler:
```suggestion
// StartImmediately sets the jobs next run to now and the startsImmediately bool to true
// This will cause the scheduler to run the job immediately when it is started
```

scheduler.go
```diff
@@ -280,10 +279,17 @@ func (j *Job) GetAt() string {
...
+func (s *Scheduler) StartImmediately() *Scheduler {
+	job := s.getCurrentJob()
+	job.nextRun = time.Now().In(s.loc)
```
JohnRoesler: Do we need to add a small buffer here in case between defining the jobs and starting the scheduler there was some small latency?

Streppel: I don't know about that... any assumption between the time of the job's definition and scheduler's start here would be incorrect since it could be of any value. By setting nextRun to now, it assumes it needs to run as soon as the next scheduler's tick occurs

JohnRoesler: 👍

PR #8, opened 2020-03-27
move tests around, rename, add more schedule func tests

### LINKED ISSUE #6: Split out tests to match files where code lives

JohnRoesler:
closes #6 

- move tests around, rename, add more schedule func tests
- actually commit go generate timeunit stringer


timeunit_string.go
```diff
@@ -0,0 +1,28 @@
+// Code generated by "stringer -type=timeUnit"; DO NOT EDIT.
```
Streppel: This is good however I don't think we use the time unit string representations anywhere... should we still keep this file?

JohnRoesler: If we don't, then we should probably remove the go generate bit from the gocron file, because I don't believe its needed for iota the timeunits

Streppel: I say we remove it, then; agree?

scheduler.go
```diff
@@ -60,8 +60,8 @@ func (s *Scheduler) Less(i, j int) bool {
 	return s.jobs[j].nextRun.Unix() >= s.jobs[i].nextRun.Unix()
 }
 
-// ChangeLoc changes the default time location
-func (s *Scheduler) ChangeLoc(newLocation *time.Location) {
+// SetLocation changes the default time location
+func (s *Scheduler) SetLocation(newLocation *time.Location) {
```
arjunmahishi: This is not backward compatible right?

Streppel: No. However, as this is a brand new repo, we are taking this chance to make some changes to the API before people begin to use it in production

arjunmahishi: :+1:

Issue #10, opened 2020-03-27
Implement an interface for time


JohnRoesler:
```
type Time interface {
    Now() time.Time
    After(d time.Duration) <-chan time.Time
    ...
}

type realTime struct{}
func (realTime) Now() time.Time { return time.Now() }
func (realTime) After(d time.Duration) <-chan time.Time { return time.After(d) }

type fakeTime struct{}
func (fakeTime) Now() time.Time { return ... }
func (fakeTime) After(d time.Duration) <-chan time.Time { return ... }
```

arjunmahishi:
Do we also want `NewTicker` to be a part of this interface?

JohnRoesler:
For sure, i was just providing an example of how I think it should look. Probably more funcs we'd need to add 👍

Streppel:
merged #12!

PR #12, opened 2020-03-28
add timeHelper interface to avoid direct coupling with the time package

arjunmahishi:
Addressing #10 

Created a `timeHelper` interface to wrap the functions we are using from the `time` package. Not wrapping the constants like `time.Sunday` because I guess that's not necessary. 

We might not need to define a `mockTime` explicitly if we use a mocking tool like [Mockary](https://github.com/vektra/mockery).

### DIFF
--- a/gocron.go
+++ b/gocron.go
@@ -48,7 +48,6 @@ const (
 
 var (
 	locker Locker
-	th     = newTimeHelper()
 )
 
 // SetLocker sets a locker implementation
--- a/gocron.go
+++ b/gocron.go
@@ -48,6 +48,7 @@ const (
 
 var (
 	locker Locker
+	th     = newTimeHelper()
 )
 
 // SetLocker sets a locker implementation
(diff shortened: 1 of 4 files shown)


arjunmahishi:
The tests are passing locally for me. Not sure why it failed here

Streppel:
@arjunmahishi unfortunately our tests are non deterministic, having a time interface would help exactly on that :)

gocron.go
```diff
@@ -48,6 +48,7 @@ const (
 
 var (
 	locker Locker
+	th     = newTimeHelper()
```
Streppel: I believe this would be better injected in the scheduler instead of being a global instance

arjunmahishi: Sure. Makes sense. Will update it in some time

arjunmahishi: `job` also uses the time functions. So, injecting it there as well.

JohnRoesler:
Awesome. Next step - mock time, so we can make all our tests better

Issue #14, opened 2020-03-29
Investigate better options for parsing time in Scheduler At func


JohnRoesler:
There should be a better way than taking in a string and parsing time

arjunmahishi:
A better way to parse time would be to use the time layout `15:04:05` and/or `15:04` with `time.Parse()` and then 
```go
parsedTime , err := time.Parse("15:04:05", t)
return parsedTime.Hour(), parsedTime.Minute(), parsedTime.Second(), nil
```
We could use regex to check the format and pick the correct template.

arjunmahishi:
I guess we can close this now.

PR #15, opened 2020-03-29
Update all exported comments and start example_test.go for godoc

JohnRoesler:
This is explains godocs pretty helpfully: https://godoc.org/github.com/fluhus/godoc-tricks

What the examples look like:
image

### DIFF
--- a/gocron.go
+++ b/gocron.go
@@ -26,7 +26,6 @@ var (
 	ErrParamsNotAdapted     = errors.New("the number of params is not adapted")
 	ErrNotAFunction         = errors.New("only functions can be schedule into the job queue")
 	ErrPeriodNotSpecified   = errors.New("unspecified job period")
-	ErrParameterCannotBeNil = errors.New("nil paramaters cannot be used with reflection")
 	ErrNotScheduledWeekday  = errors.New("job not scheduled weekly on a weekday")
 )
 
--- a/gocron.go
+++ b/gocron.go
@@ -4,14 +4,6 @@
 // for configuration. Schedule lets you run Golang functions periodically
 // at pre-determined intervals using a simple, human-friendly syntax.
 //
-// Inspired by the Ruby module clockwork <https://github.com/tomykaira/clockwork>
-// and
-// Python package schedule <https://github.com/dbader/schedule>
-//
-// See also
-// http://adam.heroku.com/past/2010/4/13/rethinking_cron/
-// http://adam.heroku.com/past/2010/6/30/replace_cron_with_clockwork/
-//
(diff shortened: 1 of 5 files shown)


job.go
```diff
@@ -110,13 +101,21 @@ func (j *Job) periodDuration() (time.Duration, error) {
...
 func (j *Job) NextScheduledTime() time.Time {
 	return j.nextRun
 }
 
-// GetWeekday returns which day of the week the job will run on
-// This should only be used when .Weekday(...) was called on the job.
-func (j *Job) GetWeekday() time.Weekday {
-	return j.startDay
+// GetAt returns the specific time of day the Job will run at
+func (j *Job) GetAt() string {
```
arjunmahishi:
```suggestion
func (j *Job) GetScheduledTime() string {
```

job.go
```diff
@@ -110,13 +101,21 @@ func (j *Job) periodDuration() (time.Duration, error) {
...
+// NextScheduledTime returns the time of the Job's next scheduled run
 func (j *Job) NextScheduledTime() time.Time {
 	return j.nextRun
 }
 
-// GetWeekday returns which day of the week the job will run on
-// This should only be used when .Weekday(...) was called on the job.
-func (j *Job) GetWeekday() time.Weekday {
-	return j.startDay
+// GetAt returns the specific time of day the Job will run at
```
arjunmahishi:
```suggestion
// GetScheduledTime returns the specific time of day the Job will run at
```

job.go
arjunmahishi: Would returning `time.Time` instead of `string` be better here?

gocron.go
```diff
@@ -20,16 +12,27 @@ package gocron
...
 	"runtime"
 	"strconv"
 	"strings"
 )
 
-// Locker provides a method to lock jobs from running
-// at the same time on multiple instances of gocron.
-// You can provide any locker implementation you wish.
+// Error declarations for gocron related errors
+var (
```
arjunmahishi: `ErrTimeFormat`, `ErrParamsNotAdapted`, `ErrNotAFunction`, `ErrPeriodNotSpecified` and `ErrParameterCannotBeNil` are unused

job.go
JohnRoesler: I agree, time would be better. I'm going to leave that for #14

JohnRoesler: I like the rename, I'll update

Streppel:
@arjunmahishi made some good points but I think this looks good now

JohnRoesler:
Agreed - great comments - one will be addressed in another issue / PR

PR #25, opened 2020-04-10
creates StartAsync and StartBlocking methods

Streppel:


scheduler.go
```diff
@@ -24,8 +24,13 @@ func NewScheduler(loc *time.Location) *Scheduler {
 	}
 }
 
-// Start all the pending Jobs using a per second ticker
-func (s *Scheduler) Start() chan struct{} {
+// StartBlocking starts all the pending jobs using a second-long ticker and blocks current thread
```
JohnRoesler, line 27:
```suggestion
// StartBlocking starts all the pending jobs using a second-long ticker and blocks the current thread
```

scheduler_test.go
```diff
@@ -90,7 +90,7 @@ func TestAt(t *testing.T) {
 	nextRun := dayJob.NextScheduledTime()
 	assert.Equal(t, expectedStartTime, nextRun)
 
-	sStop := s.Start()
+	sStop := s.StartAsync()
```
JohnRoesler, line 93:
```suggestion
	stop := s.StartAsync()
```

scheduler_test.go
```diff
@@ -90,7 +90,7 @@ func TestAt(t *testing.T) {
...
 	<-dayJobDone // Wait job done
 	close(sStop)
```
JohnRoesler, line 95:
```suggestion
	close(stop)
```

PR #33, opened 2020-04-26
Add Feature Delete Job By Tag

### LINKED ISSUE #31: How remove job?

If we send job by api, how delete old job? delete by *task object is not work,

I think we can add Id or tag ,for add job repeat,and can delete by tag/id

We can set job tag/id,and this can use this can find job,and handler job by this

auzty:
### What does this do?

- Add Feature to Create Job with Tag 
- Add Feature to Remove Job By Tag

`Resolves #31 `

- [ ] Updated `scheduler.go`
- [ ] Updated `scheduler_test.go`
- [ ] Updated `README.md`


scheduler.go
```diff
@@ -209,6 +210,25 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
 	}
 }
 
+// Remove First found Job by Tag
+func (s *Scheduler) RemoveJobByTag(tag string) {
```
arjunmahishi, line 214: Return `error`?
If the job was not found (`jobindex` < 0), the client using this will never know

scheduler.go
```diff
@@ -209,6 +210,25 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
...
+	jobindex := s.findJobsIndexByTag(tag)
+	// Remove Only if return valid index
+	if jobindex >= 0 {
+		s.jobs = removeAtIndex(s.jobs, jobindex)
+	}
+}
+
+// Find first job index by given string
+func (s *Scheduler) findJobsIndexByTag(tag string) int {
```
arjunmahishi, line 223: This function could return `(int, error)` instead of just `int`
So, if the job is found: 
```go
return i, nil
```
else:
```go
return -1, ErrJobNotFound // Exported error
```

scheduler.go
auzty: oh i see, so it' will not confuse the client who using this lib. thank you

scheduler.go
auzty: ohh ok, it will made the code clearer, thank you for the suggestion.

scheduler.go
```diff
@@ -210,23 +210,24 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
...
-	jobindex := s.findJobsIndexByTag(tag)
+// RemoveJobByTag will Remove Jobs by Tag
+func (s *Scheduler) RemoveJobByTag(tag string) error {
+	jobindex, err := s.findJobsIndexByTag(tag)
 	// Remove Only if return valid index
-	if jobindex >= 0 {
+	if err == nil && jobindex >= 0 {
 		s.jobs = removeAtIndex(s.jobs, jobindex)
 	}
+	return err
```
arjunmahishi: The standard "Golang" way would be this
```suggestion
	if err != nil {
		return err
	}
	s.jobs = removeAtIndex(s.jobs, jobindex)
```

arjunmahishi:
Apart from that, looks good to me.

scheduler.go
auzty: thank you, actually i use that logic, but idk why i so paranoid about what if somehow the return always remove at that index so i decide to move the logic like this. but i think it's just my overthinking, I am very okay with your suggestion. Thank you

PR #39, opened 2020-05-21
fixes 38

### LINKED ISSUE #38: [BUG] - StartAsync creates infinite schedulers
### Describe the bug

Calling `StartAsync` repeatedly causes a mem leak due to the neverstopping ticker and causes many concorrent runs on RunPending.

Streppel:
fixes #38

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -259,14 +259,16 @@ func (s *Scheduler) Scheduled(j interface{}) bool {
 	return false
 }
 
-// Clear deletes all Jobs
+// Clear clear all Jobs from this scheduler
 func (s *Scheduler) Clear() {
 	s.jobs = make([]*Job, 0)
 }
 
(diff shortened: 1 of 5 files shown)


scheduler.go
```diff
@@ -32,22 +38,26 @@ func (s *Scheduler) StartBlocking() {
...
+			case <-s.stopChan:
 				ticker.Stop()
+				s.running = false
 				return
 			}
 		}
 	}()
 
-	return stopped
+	return s.stopChan
```
arjunmahishi, line 60: (Unrated to this PR) Do we really need to expose this `chan` to the user? We already have a `s.Stop()` that takes care of stopping the scheduler.

Streppel: You're absolutely right, but as this would a breaking change I decided to leave it for another PR

PR #46, opened 2020-07-11
Fix removeByCondition get "slice bounds out of range"

### LINKED ISSUE #45: [BUG] - removeByCondition get "slice bounds out of range"
### Describe the bug

Loop in `removeByCondition` (scheduler.go) will keep iterating by the original value because the range loop copies the values from the slice to a local variable n; updating n will not affect the slice. It will result in panic because `slice bounds out of range` when the index of iteration goes beyond the current length of `s.jobs`

### To Reproduce

1. Run example code
2. Will result in this error: 
```
2020-07-11 08:00:11 +0000 UTC
panic: runtime error: slice bounds out of range [11:10]

goroutine 1 [running]:
github.com/go-co-op/gocron.removeAtIndex(...)
	/home/ridwanakf/go/pkg/mod/github.com/go-co-op/gocron@v0.2.0/scheduler.go:248
github.com/go-co-op/gocron.(*Scheduler).removeByCondition(0xc000032100, 0xc000098da8)
	/home/ridwanakf/go/pkg/mod/github.com/go-co-op/gocron@v0.2.0/scheduler.go:218 +0x23b
github.com/go-co-op/gocron.(*Scheduler).Remove(0xc000032100, 0x4dfda0, 0x515080)
	/home/ridwanakf/go/pkg/mod/github.com/go-co-op/gocron@v0.2.0/scheduler.go:203 +0x66
main.main()
	/home/ridwanakf/go/src/github.com/ridwanakf/personal-go-projects/cron/main.go:78 +0x1340
exit status 2
```

### Version

commit SHA: 4cc1bfb73568aeb8048dac6902e6b43df7d089ae

### Expected behavior

Should delete all expected jobs without resulting in panic

ridwanakf:
### What does this do?
Fix removeByCondition get "slice bounds out of range".

Resolves #45 

1. Adding corner case on TestRemove unit test in `schedule_test.go`
2. Changing `range` loop to traditional iteration loop on `removeByCondition` in `schedule.go` to fix slice bounds out of range.

Yes

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -213,9 +213,11 @@ func (s *Scheduler) RemoveByReference(j *Job) {
 }
 
 func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
-	for i, job := range s.jobs {
-		if shouldRemove(job) {
+	for i := 0; i < len(s.jobs); {
+		if shouldRemove(s.jobs[i]) {
 			s.jobs = removeAtIndex(s.jobs, i)
+		} else {
+			i++
 		}
 	}
 }
--- a/scheduler.go
+++ b/scheduler.go
@@ -213,13 +213,13 @@ func (s *Scheduler) RemoveByReference(j *Job) {
 }
 
 func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
-	for i := 0; i < len(s.jobs); {
-		if shouldRemove(s.jobs[i]) {
-			s.jobs = removeAtIndex(s.jobs, i)
-		} else {
-			i++
+	retainedJobs := make([]*Job, 0)
+	for _, job := range s.jobs {
+		if !shouldRemove(job) {
+			retainedJobs = append(retainedJobs, job)
 		}
 	}
+	s.jobs = retainedJobs
 }
 
 // RemoveJobByTag will Remove Jobs by Tag
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -369,11 +369,12 @@ func TestRemove(t *testing.T) {
 	scheduler := NewScheduler(time.UTC)
 	scheduler.Every(1).Minute().Do(task)
 	scheduler.Every(1).Minute().Do(taskWithParams, 1, "hello")
+	scheduler.Every(1).Minute().Do(task)
 
-	assert.Equal(t, 2, scheduler.Len(), "Incorrect number of jobs")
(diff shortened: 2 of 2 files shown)


scheduler.go
```diff
@@ -213,9 +213,11 @@ func (s *Scheduler) RemoveByReference(j *Job) {
 }
 
 func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
-	for i, job := range s.jobs {
-		if shouldRemove(job) {
+	for i := 0; i < len(s.jobs); {
+		if shouldRemove(s.jobs[i]) {
```
arjunmahishi, line 217: Will `s.jobs[i]` not panic?
Because we are iterating over `len(s.jobs)` and removing jobs in the same loop.

Maybe we should maintain a separate array of jobs that are not removed and finally assign this new array to `s.jobs`:

```go
retainedJobs := []<job-type>
if !shouldRemove(s.jobs[i]) {
    retainedJobs = append(retainedJobs, s.jobs[i])
}
s.jobs = retainedJobs
```
This would perhaps be safer.

arjunmahishi: What do you guys think?
@JohnRoesler @Streppel

ridwanakf: So, we're not actually removing any job but instead keeping the job that shouldn't be removed and assigning the new slice to `s.jobs`?

arjunmahishi: Yes.

ridwanakf: Okay, should I wait for other reviewers or change it right now?

arjunmahishi: Let's hear what the others have to say

Streppel: Thank you @ridwanakf for your contribution and @arjunmahishi for reviewing and pinging us out! 

I think @arjunmahishi  comment is absolutely valid and a good enhancement for the project. We should avoid mutate slices while iterating over them. Can you make this change @ridwanakf ?

ridwanakf: Sure, I'll work on that. btw, should I just revert back the loop using `range` since I think it's rather cleaner and also not really the problem anymore since we just keeping track of new slice?

Streppel: Feel free to do so :)

ridwanakf: Nevermind, I think it doesn't matter since we will be assigning `s.jobs` to new slice anyway. CMIIW

ridwanakf: I've made the changes, please kindly review it again.

Issue #43, opened 2020-06-15
[FEATURE] - Monthly Schedule


iambudi:
I could not find a way to schedule cron job every month on specific day (Jan 5, Feb 5, March 5 so on).

`gocron.Every(1).Month().On(5).At("hh:mm")`

Is that already possible now or should be requested as a new feature?

Thanks

iambudi:
I'm thinking doing workaround by running on Every(1).Day and check manually if the day is 5 and do the task.

JohnRoesler:
Correct, `Month` isn't a feature today. I think it would be a good feature to add 👍  I think `Months` as well would be added for every other month ability.

arjunmahishi:
Working on this. Should we just consider a `Month` as 30 days?

PR #47, opened 2020-07-18
Add support for executing jobs every Month/Months

arjunmahishi:
### What does this do?
Add support for executing jobs every Month/Months

#43 

Doesn't break any current functionality

Yes

- [x] Updated `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,8 @@ func main() {
     s2.Every(2).Days().Do(task)
     s2.Every(1).Week().Do(task)
     s2.Every(2).Weeks().Do(task)
+    s2.Every(1).Month(time.Now().Day()).Do(task)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -54,6 +54,8 @@ func main() {
     s2.Every(2).Days().Do(task)
     s2.Every(1).Week().Do(task)
     s2.Every(2).Weeks().Do(task)
+    s2.Every(1).Month(time.Now().Day()).Do(task)
+    s2.Every(2).Months(15).Do(task)
```
JohnRoesler, line 58: I think we should support At() for the month - otherwise I assume it runs at midnight by default?

arjunmahishi: Yes by default it runs at `00:00:00`. And It does support `At()`

PR #54, opened 2020-08-29
refact: rewriting scheduler logic

### LINKED ISSUE #52: [BUG] - Runs task twice
Hi everyone, I am trying to integrate `gocron` in my service. But when I tried to run task it is running it twice.

```golang
package services

import (
	"fmt"
	"time"

	"github.com/go-co-op/gocron"
)

func task() {
	fmt.Println("I am running task.")
}
func main() {
	s1 := gocron.NewScheduler(time.UTC)
	s1.Every(1).Monday().Do(task)
	s1.StartAsync()
        select{}
}
```
```
Output:
I am running task.
I am running task.
```

Is there something that I am missing or is it a bug?

Thanks you very much.

Streppel:
### What does this do?
Time scheduling was spread between job registration and actual schedule calculation, which was confusing and made testing difficult. As a matter of fact, our weekdays logic wasn't working for some edge cases and we probably had some bugs out on .Month() calculation too, although I didn't check it explicitly.

This PR rewrites our next-run scheduling logic and cover it with different tests use cases, taking in consideration weekday schedules, weekdays schedules starting immediately and exact-time schedules.

Sorry for it being big. Most of the new lines are tests cases.
 
This might fix #52. I cannot actually say this fixes it since I was not able to reproduce it, but nevertheless this PR fixes weekday logic.

None

Lots of, but might have missed something. Please bring it up if you can think of anything.

### DIFF
--- a/job.go
+++ b/job.go
@@ -7,21 +7,21 @@ import (
 
 // Job struct stores the information necessary to run a Job
 type Job struct {
-	interval             uint64                   // pause interval * unit between runs
-	unit                 timeUnit                 // time units, ,e.g. 'minutes', 'hours'...
-	periodDuration       time.Duration            // interval * unit
-	startImmediatelyFlag bool                     // if the Job should run upon scheduler start
-	jobFunc              string                   // the Job jobFunc to run, func[jobFunc]
-	atTime               time.Duration            // optional time at which this Job runs
-	err                  error                    // error related to Job
-	lastRun              time.Time                // datetime of last run
-	nextRun              time.Time                // datetime of next run
-	scheduledWeekday     *time.Weekday            // Specific day of the week to start on
-	dayOfTheMonth        int                      // Specific day of the month to run the job
-	funcs                map[string]interface{}   // Map for the function task store
-	fparams              map[string][]interface{} // Map for function and  params of function
-	lock                 bool                     // lock the Job from running at same time form multiple instances
-	tags                 []string                 // allow the user to tag Jobs with certain labels
+	interval          uint64                   // pause interval * unit between runs
+	unit              timeUnit                 // time units, ,e.g. 'minutes', 'hours'...
+	periodDuration    time.Duration            // interval * unit
+	startsImmediately bool                     // if the Job should run upon scheduler start
+	jobFunc           string                   // the Job jobFunc to run, func[jobFunc]
+	atTime            time.Duration            // optional time at which this Job runs
+	err               error                    // error related to Job
+	lastRun           time.Time                // datetime of last run
+	nextRun           time.Time                // datetime of next run
+	scheduledWeekday  *time.Weekday            // Specific day of the week to start on
+	dayOfTheMonth     int                      // Specific day of the month to run the job
+	funcs             map[string]interface{}   // Map for the function task store
+	fparams           map[string][]interface{} // Map for function and  params of function
+	lock              bool                     // lock the Job from running at same time form multiple instances
+	tags              []string                 // allow the user to tag Jobs with certain labels
 }
 
 // NewJob creates a new Job with the provided interval
--- a/job.go
+++ b/job.go
@@ -87,9 +87,9 @@ func (j *Job) ScheduledAtTime() string {
 
 // Weekday returns which day of the week the Job will run on and
 // will return an error if the Job is not scheduled weekly
-func (j *Job) Weekday() (*time.Weekday, error) {
+func (j *Job) Weekday() (time.Weekday, error) {
 	if j.scheduledWeekday == nil {
-		return nil, ErrNotScheduledWeekday
+		return time.Sunday, ErrNotScheduledWeekday
 	}
-	return j.scheduledWeekday, nil
+	return *j.scheduledWeekday, nil
 }
--- a/job.go
+++ b/job.go
@@ -9,7 +9,6 @@ import (
 type Job struct {
 	interval          uint64                   // pause interval * unit between runs
 	unit              timeUnit                 // time units, ,e.g. 'minutes', 'hours'...
-	periodDuration    time.Duration            // interval * unit
 	startsImmediately bool                     // if the Job should run upon scheduler start
 	jobFunc           string                   // the Job jobFunc to run, func[jobFunc]
 	atTime            time.Duration            // optional time at which this Job runs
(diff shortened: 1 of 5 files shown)


scheduler.go
```diff
@@ -306,23 +365,6 @@ func (s *Scheduler) Do(jobFun interface{}, params ...interface{}) (*Job, error)
 		}
 	}
 
-	if !j.startsImmediately {
-
-		if j.lastRun == s.time.Unix(0, 0) {
```
Streppel, line 311: this was a workaround that had impact on the job scheduling logic and go removed

scheduler.go
```diff
@@ -362,6 +404,9 @@ func (s *Scheduler) StartImmediately() *Scheduler {
 
 // shouldRun returns true if the Job should be run now
 func (s *Scheduler) shouldRun(j *Job) bool {
+	if j.neverRan() && j.startsImmediately {
```
Streppel, line 407: perhaps could avoid calling neverRan by setting startsImmediately to false always after first reschedule; currently we do it on weekday scheduling

timeHelper.go
```diff
@@ -6,19 +6,13 @@ type timeWrapper interface {
...
-func newTimeWrapper() timeWrapper {
-	return &trueTime{}
-}
-
 type trueTime struct{}
 
 func (t *trueTime) Now(location *time.Location) time.Time {
-	n := time.Now().In(location)
-	return t.Date(n.Year(), n.Month(), n.Day(), n.Hour(), n.Minute(), n.Second(), 0, location)
+	return time.Now().In(location)
```
arjunmahishi, line 15: Wasn't there a valid reason for how it was earlier? Have a look at his commit: 46d7545b76cfab41dc5616b015e08fa91e43df01

arjunmahishi (commented):
Have looked at the code so far. Haven't seen the tests yet.

timeHelper.go
JohnRoesler: I ran a per second job and for 5 minutes and it never drifted - IIRC this was related to that bug but is not itself an issue.

scheduler_test.go
```diff
@@ -82,50 +83,57 @@ func TestStartImmediately(t *testing.T) {
...
-	dayJob, _ := s.Every(1).Day().At(startAt).Do(func() {
-		dayJobDone <- true
+		})
+		s.scheduleAllJobs()
+
+		// Expected start time
+		expectedStartTime := time.Date(scheduletime.Year(), scheduletime.Month(), scheduletime.Day(), now.Hour(), now.Minute(), now.Add(3*time.Second).Second(), 0, time.UTC)
+		assert.Equal(t, expectedStartTime, job.ScheduledTime())
 	})
+	t.Run("runAt(3 seconds in the past) should run tomorrow at current time plus 3 seconds", func(t *testing.T) {
```
JohnRoesler, line 103:
```suggestion
	t.Run("runAt(3 seconds in the past) should run tomorrow at current time minus 3 seconds", func(t *testing.T) {
```

scheduler_test.go
```diff
@@ -82,50 +83,57 @@ func TestStartImmediately(t *testing.T) {
 }
 
 func TestAt(t *testing.T) {
-	s := NewScheduler(time.UTC)
+	t.Run("runAt(3 seconds in the future) should run today at current time plus 3 seconds", func(t *testing.T) {
```
JohnRoesler, line 86: the test below is called `TestAtFuture` - this should either move there, or could we combine these into a single test and make it a table test?

scheduler_test.go
```diff
@@ -535,38 +422,750 @@ func TestScheduler_Stop(t *testing.T) {
...
-	)
+	assert.Equal(t, now.Add(time.Second*3).Second(), nextRun.Second())
+}
+
+func TestScheduler_ScheduleDays(t *testing.T) {
+	sched := NewScheduler(time.UTC)
+	fakeTime := fakeTime{}
+	sched.time = &fakeTime
+
+	t.Run("schedule for days without .At(time)", func(t *testing.T) {
```
JohnRoesler, line 451: the t.Run blocks look to be repeated, could make this a [table test](https://github.com/golang/go/wiki/TableDrivenTests)?

job.go
```diff
@@ -108,9 +110,9 @@ func (j *Job) ScheduledAtTime() string {
 
 // Weekday returns which day of the week the Job will run on and
 // will return an error if the Job is not scheduled weekly
-func (j *Job) Weekday() (time.Weekday, error) {
-	if j.unit == weeks {
-		return j.startDay, nil
+func (j *Job) Weekday() (*time.Weekday, error) {
```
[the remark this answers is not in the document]
Streppel: You're right, this is not backward compatible. About the second thought, I didn't get it exactly; you meant at line 117?

scheduler_test.go
Streppel: That makes sense, will take a look at it later tonight

job.go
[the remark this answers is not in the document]
arjunmahishi:
```suggestion
func (j *Job) Weekday() (time.Weekday, error) {
```

job.go
```diff
@@ -108,9 +110,9 @@ func (j *Job) ScheduledAtTime() string {
...
+	if j.scheduledWeekday == nil {
+		return nil, ErrNotScheduledWeekday
 	}
-	return time.Sunday, ErrNotScheduledWeekday
+	return j.scheduledWeekday, nil
```
arjunmahishi, line 117:
```suggestion
	return *j.scheduledWeekday, nil
```

Streppel: I see. We'll still need the check, though. Users may call Weekday on a non-weekday job and that'd panic. An workaround for this is setting a default weekday that'll be ignored (which is our current solution actually). What do you think?

arjunmahishi: Yes, we will need the check for sure.

Streppel: done

Streppel:
@arjunmahishi @JohnRoesler cleaned up tests and fixed monthly issues, good to go now :-) there are tests missing (startImmediately reschedule etc) but we didn't have a strong test suite for these before anyway, so what do you think about leaving these tests for a future PR?

Our schedule method is now much more assertive, intelligent and readable IMO. Download the branch and let me know what you think.

arjunmahishi (approved):
LGTM

JohnRoesler (approved):

Issue #41, opened 2020-06-03
[FEATURE] - Error chan on scheduler


110:
https://github.com/jasonlvhit/gocron/issues/134

This feature was discussed before.  
hop add this  feature .

JohnRoesler:
This would be a grand addition. Something you're interested in contributing on?

ofonimefrancis:
> This would be a grand addition. Something you're interested in contributing on?

I would love to work on this @JohnRoesler, what logger would you recommend?

JohnRoesler:
Exactly - we'd want an interface that the user would them implement with their specific chosen logger! Let us know if you want to discuss or if you need any assistance! We're on slack - there is a link in the readme

arjunmahishi:
@ofonimefrancis you can use an interface like this:
```go
type Logger interface {
	Debugf(format string, args ...interface{})
	Infof(format string, args ...interface{})
	Printf(format string, args ...interface{})
	Warnf(format string, args ...interface{})
	Warningf(format string, args ...interface{})
	Errorf(format string, args ...interface{})
	Fatalf(format string, args ...interface{})
	Panicf(format string, args ...interface{})

	Debug(args ...interface{})
	Info(args ...interface{})
	Print(args ...interface{})
	Warn(args ...interface{})
	Warning(args ...interface{})
	Error(args ...interface{})
	Fatal(args ...interface{})
	Panic(args ...interface{})
}
```
(Partly taken from logrus)

You can remove some of the methods like `Panic`, `Print` to keep it simple.

arjunmahishi:
I am planning to work on this. Will mostly create a PR sometime today. @ofonimefrancis, let me know if you have already started working on it

arjunmahishi:
@JohnRoesler @Streppel Can we list down the things we want to log with their log-levels?

Here are some things I can think of:
```shell
# Job related
[DEBUG] Running job: <job name>
[INFO] Ran job: <job name>
[ERROR] Failed to run job: <job name>

# Scheduler related
[INFO] Scheduled the job '<job name>' to run at '<time>'
```

arjunmahishi:
I've created a simple custom logging framework and pushed it to the branch [feature/custom-logger](https://github.com/go-co-op/gocron/tree/feature/custom-logger). Let me know what you guys think.

Streppel:
@arjunmahishi Cool! Will take a look at it this weekend (been in a hurry this week)

arjunmahishi:
[...]

Issue #50, opened 2020-07-24
[BUG] - Lock service is not working


ldegaetano:
Hi every one, i was checking that the feature for preventing run a task twice using lock service is not working at all. In the function   `func (s *Scheduler) run(job *Job) error`  at line 180, when call to `locker.Lock(key)`  the response is ignored, so if `lock()`  return false the job can be executed.

```
func (s *Scheduler) run(job *Job) error {
	if job.lock {
		if locker == nil {
			return fmt.Errorf("trying to lock %s with nil locker", job.jobFunc)
		}
		key := getFunctionKey(job.jobFunc)

		locker.Lock(key)
		defer locker.Unlock(key)
	}

	job.lastRun = s.time.Now(s.loc)
	go job.run()

	return nil
}
```

Is there a scenario I am not seeing or is it a bug?

Thanks you very  much.

maxsuelmarinho:
It started to ignore the return of the `locker.Lock(key)` in this [pull request #151](https://github.com/jasonlvhit/gocron/pull/151/files#diff-0ab145ca0ab7d0d679361c0f6263a49aR63). 
As we can see the @Streppel's comment: 
"will change our locker signature in the future to remove errors just like Sync.Locker", which is this one:
https://golang.org/pkg/sync/#Locker

`type Locker interface {
    Lock()
    Unlock()
}`

But it indeed prevents previous Locker implementations to work properly. For instance, the [lock example using redis](https://github.com/go-co-op/gocron/blob/master/example/lock.go#L40) is not working anymore.

Streppel:
Indeed, this is critical. @arjunmahishi do you think you can take a look at this? I'm still reviewing the log PR and fixing some stuff on our scheduler V2

JohnRoesler:
Hey all, we have since removed the lock implementation from gocron and are recommending the jobs themselves handle locking in the event that is needed. This is in our readme now:

>  * Q: I'm running multiple pods on a distributed environment. How can I make a job not run once per pod causing duplication? 
> * A: We recommend using your own lock solution within the jobs themselves (you could use [Redis](https://redis.io/topics/distlock), for example)

PR #57, opened 2020-09-09
Custom logger

arjunmahishi:
- Create a new Logger interface with debug, info, and error
      levels
- Introduce a new argument to NewScheduler to accept custom
      options from the user

#41 

`NewScheduler` now accepts custom options from the user

### DIFF
--- a/logger.go
+++ b/logger.go
@@ -12,6 +12,14 @@ type Logger interface {
 	Error(args ...interface{})
 }
 
+// OptionLogger returns the SchedulerOption for
+// for passing in a custom logger to NewScheduler
+func OptionLogger(logger Logger) SchedulerOption {
+	return func(s *Scheduler) {
+		s.logger = logger
+	}
+}
+
 func (s *Scheduler) initLogger(l Logger) {
 	s.logger = l
 }
--- a/logger.go
+++ b/logger.go
@@ -1,0 +1,55 @@
+package gocron
+
+type Logger interface {
+	Debugf(format string, args ...interface{})
+	Infof(format string, args ...interface{})
+	Errorf(format string, args ...interface{})
+
+	Debug(args ...interface{})
+	Info(args ...interface{})
+	Error(args ...interface{})
+}
+
+var (
+	logger Logger
+)
+
+func initLogger(l Logger) {
+	logger = l
+}
+
+func debugf(format string, args ...interface{}) {
+	if logger != nil {
+		logger.Debugf(format, args...)
+	}
+}
+
+func debug(args ...interface{}) {
+	if logger != nil {
+		logger.Debug(args...)
+	}
+}
+
+func infof(format string, args ...interface{}) {
+	if logger != nil {
+		logger.Infof(format, args...)
+	}
+}
+
+func info(args ...interface{}) {
+	if logger != nil {
+		logger.Info(args...)
+	}
+}
+
+func errf(format string, args ...interface{}) {
+	if logger != nil {
+		logger.Errorf(format, args...)
+	}
+}
+
+func err(args ...interface{}) {
+	if logger != nil {
+		logger.Error(args...)
+	}
+}
--- a/logger.go
+++ b/logger.go
@@ -1,5 +1,7 @@
 package gocron
 
+// Logger interface the custom logger should
+// conform to
 type Logger interface {
 	Debugf(format string, args ...interface{})
 	Infof(format string, args ...interface{})
--- a/logger.go
+++ b/logger.go
@@ -12,46 +12,42 @@ type Logger interface {
 	Error(args ...interface{})
 }
 
-var (
-	logger Logger
-)
-
-func initLogger(l Logger) {
-	logger = l
+func (s *Scheduler) initLogger(l Logger) {
+	s.logger = l
 }
 
-func debugf(format string, args ...interface{}) {
-	if logger != nil {
-		logger.Debugf(format, args...)
+func (s *Scheduler) debugf(format string, args ...interface{}) {
+	if s.logger != nil {
+		s.logger.Debugf(format, args...)
 	}
 }
 
-func debug(args ...interface{}) {
-	if logger != nil {
-		logger.Debug(args...)
+func (s *Scheduler) debug(args ...interface{}) {
+	if s.logger != nil {
+		s.logger.Debug(args...)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


scheduler.go
```diff
@@ -570,3 +571,15 @@ func (s *Scheduler) scheduleAllJobs() {
...
 }
+
+func handleOptions(options ...interface{}) {
+	for _, option := range options {
+		switch option.(type) {
+		case Logger:
+			// Note: The logger will only be initilized if the Logger
+			// interface is conformed to. The logs will not be printed
+			// in this case
+			initLogger(option.(Logger))
```
Streppel, line 582: shouldn't the logger be attached to a scheduler instance? considering a log struct may hold state itself as well

arjunmahishi: Makes sense!
Instead of making the `logger` global, I'll make it a member of `Scheduler`. In which case, the `handleOptions` function can also become a method of `Scheduler` right?

logger.go
```diff
@@ -0,0 +1,53 @@
+package gocron
+
+// Logger interface the custom logger should
+// conform to
+type Logger interface {
```
JohnRoesler, line 5: this is real cool. have you thought about if it makes sense to have other standard log levels? warn, trace, e.g.

scheduler.go
Streppel: a good approach to optional parameters that I use [looks like this](https://stackoverflow.com/a/26326418/1670308)

logger.go
arjunmahishi: Yes... But I was thinking, this interface needs to be as simple and generic as possible. 
But also, once we release this, adding any new method to this interface would not be backward compatible. So, what do you think? Should I add all the other levels?

```go
type Logger interface {
	Debugf(format string, args ...interface{})
	Infof(format string, args ...interface{})
	Printf(format string, args ...interface{})
	Warnf(format string, args ...interface{})
	Warningf(format string, args ...interface{})
	Errorf(format string, args ...interface{})
	Fatalf(format string, args ...interface{})
	Panicf(format string, args ...interface{})

	Debug(args ...interface{})
	Info(args ...interface{})
	Print(args ...interface{})
	Warn(args ...interface{})
	Warning(args ...interface{})
	Error(args ...interface{})
	Fatal(args ...interface{})
	Panic(args ...interface{})
}
```

scheduler.go
arjunmahishi: :100: This is really cool. Will implement this. Thanks!

logger.go
Streppel: This is just an idea: from what I can see the current idea is the user to pass a log that satisfies this interface so we can use it internally. What if we, instead of accepting this logger interface, accepted an io.Writer to know where to write data to. This way we would break this dependency and it'd be pretty easy to test too. Let's stress this further if it makes sense.

arjunmahishi: Interesting thought. But the user will not have any control over the format of the log. Some people follow a specific structure in their logs, right?

JohnRoesler: Taking a step back for a moment - what are we going to be logging? Errors primarily i would assume.

What if we add a channel to our scheduler or the job itself that we return job run errors on for the caller to handle however they so choose. Instead of adding any logging in the package itself?

Streppel: If we stop to think about it, errors on schedule. Anything else can be caught up by closures from within the job function the client passes itself... So I'm starting to question actually the real need for this indeed. When (if) we move from our own scheduling logic to using a time.Ticker/After approach we won't even need to worry about this anymore. What do you think @arjunmahishi?

arjunmahishi: Makes sense... There is no real need for this, agreed.

arjunmahishi: So, are we going with the error channel approach?
Or does this need more thought?

arjunmahishi: @Streppel @JohnRoesler Waiting to hear your thoughts

Streppel: @arjunmahishi sorry! Pretty busy last couple of days and github emails got lost in my inbox. I'd say to keep this branch open while we stress this idea a little bit further. We could try to identify within the code what could go wrong in places that users wouldn't be able to handle by themselves.

One example that comes to mind [is this one](https://github.com/go-co-op/gocron/blob/04be48f61037ca7e035c0f17f4547c3b0ba93a0e/scheduler.go#L254), which could happen if a job fails to run or to schedule. We are safe against scheduling failures as we do all the checks on the job creation. And the running part fails if the scheduler is supposed to lock the job on a distributed system without a locker implementation ([something we know that isn't working at the moment and needs to be fixed](https://github.com/go-co-op/gocron/issues/50), and after the fix we could fail to start a job under such conditions, or maybe consider a "locking job"  if it has locker instead of doing so with a flag). So, in the end, I believe this error isn't worth considering.

And actually, if we look at it... Scheduling and running jobs is basically everything this lib needs to do. Could you guys point to any other places that you think would need extra logging?

Streppel: (btw fellow comrades let me know if any of you want to get the linked issue, otherwise I'll take it)

arjunmahishi:
@JohnRoesler @Streppel 
Let's close this PR. The error channel seems like a better idea to notify the user of failures. Plus, this will also let the user take appropriate action. Let me see I can put something together and make another PR

logger.go
arjunmahishi: Currently, these are the functions whose errors we are not handling
```
scheduler.runAndReschedule
locker.Lock
locker.Unlock
callJobFuncWithParams
```

Streppel: Afterall, I don't think we should be handling logs internally. I believe callJob and runAndReschedule shouldn't need to return errors, and if something goes wrong with the distrubuted locker, users would probably want a little more control other than logging on how to handle the error; agree? @arjunmahishi

arjunmahishi: Yes. The user should be able to handle the error.

arjunmahishi: So, would the error channel that John suggested be useful here?

The scheduler could have a `func (s *sched) CatchErrors() chan` method to expose the error channel.

Streppel: I believe that'd be the way to go! Exposing though a Errors() method or something along those lines, making sure the program doesn't freeze if the buffer gets filled

arjunmahishi: OK then! Will close this PR and work a new one :)

Issue #60, opened 2020-09-15
[BUG] - All Jobs Start Immediately


graywolf336:
### Describe the bug
I have three jobs which I want to run at a specific time of day and a specific day of the month. None of them have `StartImmediately()` so I would not expect them to run whenever they get scheduled.

### To Reproduce

1. `scheduler.Every(1).Day().At("00:30").Do()`
1. Expect the job to not run
1. See the logs of the job running

### Version
`v0.3.1` from `go.mod` file

### Expected behavior
The expected behavior is that because I do not have `StartImmediately()` attached to it they would not run immediately at the start up and instead would only run at the times scheduled.

### Additional context
![Screen Shot 2020-09-15 at 8 17 09 AM](https://user-images.githubusercontent.com/850391/93215756-7a422b80-f72c-11ea-982a-40263ecd1b88.png)

arjunmahishi:
Hey @graywolf336, 
I am not able to reproduce this with the steps you have mentioned. Is it possible for you to paste the exact snippet of code here?

This is working fine for me:
```go
func main() {
    s := gocron.NewScheduler(time.Local)
    s.Every(3).Seconds().At("20:35").Do(func() {
        log.Println("DONE")
    })  
    log.Println("Starting..")
    <-s.StartAsync()
}
```
Output:
```bash
 $ go run main.go 
2020/09/15 20:46:02 Starting..
2020/09/15 20:46:05 DONE
2020/09/15 20:46:08 DONE
2020/09/15 20:46:11 DONE
```

The same works for `Day()`, `Week()` and `Month()` also

arjunmahishi:
Hey @graywolf336, 
I am not able to reproduce this with the steps you have mentioned. Is it possible for you to paste the exact snippet of code here?

This is working fine for me:
```go
func main() {
    s := gocron.NewScheduler(time.Local)
    s.Every(3).Seconds().Do(func() {
        log.Println("DONE")
    })  
    log.Println("Starting..")
    <-s.StartAsync()
}
```
Output:
```bash
 $ go run main.go 
2020/09/15 20:46:02 Starting..
2020/09/15 20:46:05 DONE
2020/09/15 20:46:08 DONE
2020/09/15 20:46:11 DONE
```

The same works for `Day()`, `Week()` and `Month()` also

graywolf336:
Ah, I am starting the scheduler and then adding the jobs to it after the start was done.

```go
package main

import (
	"log"
	"os"
	"os/signal"
[...]

Issue #58, opened 2020-09-10
scheduleNextRun when the job create after schedule started


Chenghu-w:
job run immediately when I remove the job and create job again after schedule started。

expect:
scheduleNextRun when the job create after schedule started
or
add new funcion 'UpdateInterval'

testing:

```go
package main

import (
	"fmt"
	"time"

	"github.com/go-co-op/gocron"
)

var (
	lastTime time.Time
	s        *gocron.Scheduler
)

func task() {
	now := time.Now()
	fmt.Println(now.Sub(lastTime).Seconds())
	lastTime = now
	s.Remove(task)
	s.Every(3).Seconds().Do(task)
}

func main() {
	s = gocron.NewScheduler(time.UTC)
	_close := s.StartAsync()

	lastTime = time.Now()
	s.Every(3).Seconds().Do(task)

	time.AfterFunc(time.Duration(time.Second*10), func() { s.Stop() })

	<-_close
}
```

Streppel:
@Chenghu-w thanks for reporting! this has been fixed now :) please report back here if you see anything usual

apremalal:
@Streppel thanks for the fix. Could you release a new tag with the fix?

Streppel:
@apremalal done!

apremalal:
@Streppel thanks and appreciate your help

PR #61, opened 2020-09-20
fix: scheduler schedules jobs registration when already running

Streppel:
### What does this do?
scheduler schedules jobs registration when already running

#60 #58


arjunmahishi (approved):
LGTM

JohnRoesler (approved):

PR #62, opened 2020-09-21
refact: removing legacy deterministic tests

Streppel:
### What does this do?
These cases are already covered on TestScheduler_FirstSchedule. Removing these tests due cleans up the test file and makes our lives better since they were deterministic (they failed as a false positive during specific times o the day)

none


arjunmahishi (approved):

JohnRoesler:
Yay!

PR #63, opened 2020-09-21
Add option to limit maximum concurrent jobs

### LINKED ISSUE #53: [FEATURE] - Maximum concurrent jobs

When running resource intensive jobs, it may be desirable to limit how many are executed at once.

I would like to be able to configure maximum running jobs.

I can probably do a hacky implementation of this with the Locker feature.

dustin-decker:
### What does this do?
This feature is backwards compatible and allows limiting the maximum concurrent jobs.
It is useful when running resource intensive jobs and a precise start time is not critical.

Resolves #53

None

No

- [X] Updated `example_test.go`
- [X] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -1,12 +1,15 @@
 package gocron
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"reflect"
 	"sort"
 	"strings"
 	"time"
+
+	"golang.org/x/sync/semaphore"
 )
 
 // Scheduler struct stores a list of Jobs and the location of time Scheduler
@@ -19,6 +22,8 @@ type Scheduler struct {
 	stopChan chan struct{} // signal to stop scheduling
 
 	time timeWrapper // wrapper around time.Time
+
+	sem *semaphore.Weighted
 }
 
 // NewScheduler creates a new Scheduler
@@ -30,6 +35,12 @@ func NewScheduler(loc *time.Location) *Scheduler {
 		stopChan: make(chan struct{}),
 		time:     &trueTime{},
 	}
+}
+
+// SetMaxConcurrentJobs limits how many jobs can be running at the same time.
+// This is useful when running resource intensive jobs and a precise start time is not critical.
+func (s *Scheduler) SetMaxConcurrentJobs(n int) {
+	s.sem = semaphore.NewWeighted(int64(n))
 }
 
 // StartBlocking starts all the pending jobs using a second-long ticker and blocks the current thread
@@ -264,6 +275,10 @@ func (s *Scheduler) runAndReschedule(job *Job) error {
 }
 
 func (s *Scheduler) run(job *Job) error {
+	if s.sem != nil {
+		s.sem.Acquire(context.Background(), 1)
+	}
+
 	if job.lock {
 		if locker == nil {
 			return fmt.Errorf("trying to lock %s with nil locker", job.jobFunc)
@@ -275,7 +290,13 @@ func (s *Scheduler) run(job *Job) error {
 	}
 
 	job.lastRun = s.time.Now(s.loc)
-	go job.run()
+
+	go func() {
+		job.run()
+		if s.sem != nil {
+			s.sem.Release(1)
+		}
+	}()
 
 	return nil
 }
--- a/README.md
+++ b/README.md
@@ -19,10 +19,11 @@ Examples:
 package main
 
 import (
-	"fmt"
-	"time"
+    "fmt"
+    "time"
+    "runtime"
 
-	"github.com/go-co-op/gocron"
+    "github.com/go-co-op/gocron"
 )
 
 func task() {
@@ -36,6 +37,10 @@ func taskWithParams(a int, b string) {
 func main() {
     // defines a new scheduler that schedules and runs jobs
     s1 := gocron.NewScheduler(time.UTC)
+
+    // we have very intensive jobs with flexible schedules, so limiting the
+    // maximum jobs that can run at a time
+    s1.SetMaxConcurrentJobs(runtime.NumCPU())
 
     s1.Every(3).Seconds().Do(task)
 
--- a/example_test.go
+++ b/example_test.go
@@ -23,6 +23,16 @@ func ExampleScheduler_At() {
 	s.Every(1).Monday().At("10:30:01").Do(task)
 }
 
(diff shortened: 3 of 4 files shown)


JohnRoesler (approved):
Awesome and simple

scheduler.go
```diff
@@ -275,7 +286,16 @@ func (s *Scheduler) run(job *Job) error {
 	}
 
 	job.lastRun = s.time.Now(s.loc)
-	go job.run()
+
+	if s.sem != nil {
+		s.sem.Acquire(context.Background(), 1)
```
Streppel, line 291: If the semaphore is acquired here and locked until freed, job.lastRun (line 288) will have an incorrect value. Maybe we should this check be at the start of the function?

Streppel:
Thanks for your contribution @dustin-decker! Really nice of use of weighted semaphores here. Just one little comment there but other than that, looks really good :+1:

scheduler.go
dustin-decker: Good catch. Having it in the beginning also enables another scheduler to lock the job if it's waiting to acquire the semaphore too.

Streppel (approved):

scheduler.go
dustin-decker: @Streppel, what happens if the lock is already acquired by the time it attempts to lock now? Does the goroutine leak (blocked forever) and the semaphore is never released?

scheduler.go
```diff
@@ -264,6 +275,10 @@ func (s *Scheduler) runAndReschedule(job *Job) error {
 }
 
 func (s *Scheduler) run(job *Job) error {
+	if s.sem != nil {
+		s.sem.Acquire(context.Background(), 1)
```
Streppel, line 279: Another question. Upon freezing here due to lack of weights, should this job's ryn be dropped? Otherwise they'll just get "throttled"

Streppel (changes requested):
Actually, @dustin-decker, I believe we'll need to stress this a little bit further. The way it is currently implemented, the scheduler as a whole will freeze upon running out of available spots. 

Plus, we need to decide if the jobs will either be dropped or have their runs "throttled", which will cause the jobs to lose any sync and run at random times in the order they were added, which is not a very interesting behavior. 

This needs to work nicely with the distributed locker too, making these locks only work per instance. 

Any thoughts on this?

scheduler.go
Streppel: The locker is meant for distributed computation. We'd actually need to put the semaphore after the locker check. Check my last comment at the thread

Streppel:
@farwydi I see your point. A feasible approach would be to not run the job if there's no ticket avaiable, but be rescheduled nevertheless so it can retry on its next interval.

temamagic:
@Streppel 
> A feasible approach would be to not run the job if there's no ticket avaiable, but be rescheduled nevertheless so it can retry on its next interval.

what is the difference from starting and skipping a task? It can run on next interval anyway

Streppel:
@temamagic that's basically it, no big deal indeed. Skip the running, but continue on with the scheduling logic nevertheless. Thing is, not being able to run should not be a blocking action to rescheduling, that's why we should go for TryAcquire the ticket instead of Acquire

scheduler.go
```diff
@@ -32,6 +37,12 @@ func NewScheduler(loc *time.Location) *Scheduler {
 	}
 }
 
+// SetMaxConcurrentJobs limits how many jobs can be running at the same time.
+// This is useful when running resource intensive jobs and a precise start time is not critical.
+func (s *Scheduler) SetMaxConcurrentJobs(n int) {
```
farwydi, line 42:
```suggestion
func (s *Scheduler) SetMaxConcurrentJobs(n int) error {
	if n < 1 {
		return ErrConcurrentJobNegative
	}
	s.sem = semaphore.NewWeighted(int64(n))
	return nil
}
```

farwydi (commented):
you need to add protection against incorrect values

scheduler.go
Streppel, line 279:
```suggestion
		s.sem.TryAcquire(context.Background(), 1)
		return nil
```

With that, in order to keep rescheduling working effectively, we'll need to bring this to after line 292, which isn't a problem anymore since we won't be locking

farwydi, line 279:
```suggestion
		if !s.sem.TryAcquire(1) {
			// Perhaps this is considered a error?
			return nil
		}
```

farwydi (commented):
Then the task will just be skipped

farwydi:
Here is a test that identifies the problem.

```go
func TestJobAlreadyIsRunningSingleMode(t *testing.T) {
	sched := NewScheduler(time.UTC)
	sched.SetMaxConcurrentJobs(1)
	hit := 0
	sched.Every(1).Second().Do(func(prevHit *int) {
		if *prevHit > 0 {
			assert.Fail(t, "job should not re-call")
		}
		*prevHit = *prevHit + 1
		time.Sleep(2 * time.Second)
	}, &hit)
	sched.RunPending()
	time.Sleep(1 * time.Second)
	sched.RunPending()
	time.Sleep(1 * time.Second)
	sched.RunPending()
	time.Sleep(1 * time.Second)
}
```

dustin-decker:
I appreciate the collaboration on this!
I was thinking about a configurable approach:
- Best effort strategy - run blocks until it can acquire the semaphore
- Reschedule strategy - job reschedules and tries again next time

What do you think?

scheduler.go
dustin-decker: Just accepting a uint is probably cleaner here.

Streppel:
Adding strategies might be an interesting idea. I believe using a reschedule strategy by default would be the proper way to go, though. 

Out of curiosity @dustin-decker, what kind of resource intensive jobs you had in mind when you thought about this? IO bound, CPU bound?

dustin-decker:
For my use case, mostly CPU bound. Following the schedule rigidly is not so important, and for infrequent jobs I would not want them to wait for their next scheduling.

farwydi:
Choosing a strategy adds flexibility.
@dustin-decker Best effort strategy - you need to set a limit otherwise it may lead to the accumulation of goroutines due to a failure.

JohnRoesler:
What's left to do on this one? Add the strategies concept? we're thinking `reschedule`, `wait`, others?

PR #64, opened 2020-09-23
feat: Add single mode in job

farwydi:
### What does this do?

This mode prevents double start.


Streppel:
Hey @farwydi, thanks for contributing. Let me see if I got it right: is this to prevent "overrunning" a job upon itself?

farwydi:
Yes sir, similar #63

Streppel:
Oh, cool! I believe this could make good use of [singleflight](https://godoc.org/golang.org/x/sync/singleflight), have you considered using it?

dustin-decker:
It'd be nice if there were a good way to provide feedback when a job is scheduled to run but is still running. In resource-bound cases it could be a good indicator of resource starvation. I'm not sure of the best way to do this inside of a library though.

Streppel:
@farwydi did you take a look at a possible singleflight solution? I believe it'd be the way to go here

farwydi:
I didn't really understand the meaning TestSwap 🐰

farwydi:
@Streppel @dustin-decker I think this option will help us add more subtle settings with different behavior options in the future. How do you look at it?

scheduler_test.go
```diff
@@ -252,17 +252,17 @@ func TestSwap(t *testing.T) {
 	s.Every(2).Minute().Do(task)
 
 	jb := s.Jobs()
-	var jobsBefore []Job
-	for _, p := range jb {
-		jobsBefore = append(jobsBefore, *p)
+	jobsBefore := make([]*Job, 2)
```
dustin-decker, line 255: Setting this to `2` instead of `len(jb)` is fragile.

dustin-decker (dismissed):
Please alter the fragile allocation, otherwise looks good to me.

job.go
```diff
@@ -44,11 +59,19 @@ func NewJob(interval uint64) *Job {
 
 // Run the Job and immediately reschedule it
 func (j *Job) run() {
-	callJobFuncWithParams(j.funcs[j.jobFunc], j.fparams[j.jobFunc])
-	j.runCount++
+	switch j.runConfig.mode {
+	case SingletonMode:
+		_, j.err, _ = j.limiter.Do("main", func() (interface{}, error) {
```
Streppel, line 64: Just asking for the sake of it, but is this singleton mode related to the specific job in question? Because the way it is written here it will affect other singleton mode jobs as well, as they will all await for the "main"  keyword. If that's not expected, we could change this key to something job-specific in order to identify each particular job.

Streppel: Also, I'm sorry for taking so long to respond, I've been super busy nowadays :( will try to check in here more often

farwydi: Each job has its own ```limiter singleflight.Group```

```go
func (g *Group) Do(key string, fn func() (interface{}, error)) (v interface{}, err error, shared bool) {
	g.mu.Lock()
	if g.m == nil {
		g.m = make(map[string]*call)
	}
        // here
	if c, ok := g.m[key]; ok {
```

Streppel (dismissed):

Streppel:
Cool @farwydi! I'd just say to update the README file as well

job.go
```diff
@@ -155,10 +178,15 @@ func (j *Job) Weekday() (time.Weekday, error) {
...
-	}
+	j.runConfig.finiteRuns = true
+	j.runConfig.finiteRuns = n
+}
+
+// SingletonMode Sets the mode to block startup if the current job has not finished
+func (j *Job) SingletonMode() {
+	j.Lock()
+	defer j.Unlock()
+  j.runConfig.mode = SingletonMode
```
JohnRoesler, line 189:
```suggestion
	j.runConfig.mode = SingletonMode
```

example_test.go
```diff
@@ -108,7 +108,14 @@ func ExampleJob_LimitRunsTo() {
...
 	job.LimitRunsTo(2)
-	s.StartAsync()
+	<-s.StartAsync()
+}
+
+func ExampleJob_SingletonMode() {
+	s := gocron.NewScheduler(time.UTC)
+	job, _ := s.Every(1).Second().Do(task)
+	job.SingletonMode()
+  <-s.StartAsync()
```
JohnRoesler, line 118:
```suggestion
		<-s.StartAsync()
```

example_test.go
```diff
@@ -108,7 +108,14 @@ func ExampleJob_LimitRunsTo() {
...
 	job.LimitRunsTo(2)
-	s.StartAsync()
+	<-s.StartAsync()
+}
+
+func ExampleJob_SingletonMode() {
+	s := gocron.NewScheduler(time.UTC)
+	job, _ := s.Every(1).Second().Do(task)
+	job.SingletonMode()
+		<-s.StartAsync()
```
JohnRoesler, line 118:
```suggestion
	<-s.StartAsync()
```

job.go
```diff
@@ -155,10 +178,15 @@ func (j *Job) Weekday() (time.Weekday, error) {
 func (j *Job) LimitRunsTo(n int) {
 	j.Lock()
 	defer j.Unlock()
-	j.runConfig = runConfig{
-		finiteRuns: true,
-		maxRuns:    n,
-	}
+	j.runConfig.finiteRuns = true
+	j.runConfig.finiteRuns = n
```
JohnRoesler, line 182:
```suggestion
	j.runConfig. maxRuns = n
```

job_test.go
```diff
@@ -20,6 +22,23 @@ func TestTags(t *testing.T) {
 	assert.ElementsMatch(t, j.Tags(), []string{"tags", "tag", "some"})
 }
 
+func TestSingletonMode(t *testing.T) {
```
JohnRoesler, line 25: this test itself is passing, and passes in a single test run, but when run in aggregate, the next test is failing. Seems the scheduler does in fact send out job that is then waiting on the singleflight to allow it to run, and that goroutine isn't properly stopped / canceled on `s.Stop()` - looking into this

JohnRoesler: indeed. Adding 

```diff
	s.Stop()
+	time.Sleep(2 * time.Second)
```

sleep to the end of the test, confirms that a job is run after the scheduler is stopped.

JohnRoesler: so, it appears because the scheduler is scheduling another job, and it's sent to the singleton which includes a wait group, even though the scheduler is stopped, the jobs aren't currently told about that stoppage.

I am interested in how best people view to solve this. I found *a* solution by adding a context with cancel into the job itself:

```golang
func TestSingletonMode(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	s := NewScheduler(time.UTC)
	var trigger int32
	job, _ := s.Every(1).Second().Do(func() {
		select {
		case <-ctx.Done():
			return
		default:
			if atomic.LoadInt32(&trigger) == 1 {
				t.Fatal("Restart should not occur")
			}
			atomic.AddInt32(&trigger, 1)
			fmt.Println("I am a long task")
			time.Sleep(3 * time.Second)
		}
	})
	job.SingletonMode()
	s.StartAsync()
	time.Sleep(2 * time.Second)
	cancel()
	s.Stop()
	time.Sleep(2 * time.Second)
}
```

If we go this route - we're essentially saying, you the caller are responsible for making sure your job doesn't run again if you stop the scheduler.

I don't know that there is a way to build it into the library because we can't force jobs to take in a context that we can cancel, or even that the job has to stop when it's cancelled.

JohnRoesler:
This has been merged in #123

PR #65, opened 2020-09-25
fix on skipping job running when job.lock enabled

### LINKED ISSUE #50: [BUG] - Lock service is not working

Kemdole:
### What does this do?
If the cron has lock implementation(e.g. Redis example), once the "success" return false the job execution will be skipped.

Resolves #50 

None

No

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -270,7 +270,15 @@ func (s *Scheduler) run(job *Job) error {
 		}
 		key := getFunctionKey(job.jobFunc)
 
-		locker.Lock(key)
+		success, err := locker.Lock(key)
+		if err != nil {
+			return err
+		}
+		if !success {
+			job.lastRun = s.time.Now(s.loc)
+			return nil
+		}
+
 		defer locker.Unlock(key)
 	}
 


scheduler.go
```diff
@@ -270,7 +270,15 @@ func (s *Scheduler) run(job *Job) error {
 		}
 		key := getFunctionKey(job.jobFunc)
 
-		locker.Lock(key)
+		success, err := locker.Lock(key)
+		if err != nil {
+			return err
+		}
+		if !success {
+			job.lastRun = s.time.Now(s.loc)
```
Streppel, line 278: Should we update lastRun here? The job didn't actually run here

AlexMin314: Hmmm ya. I am also quite concerned about the line when I put it. The idea was because the unsuccessful job by locking will be regarded as a skipped job, I thought the line let the job be re-scheduled for the next interval. Please let me know the approach is not good. :)

Kemdole:
Or it will be better to encourage to implement the locking approach within job functions in the documentation and remove the not working example?

Streppel:
@Kemdole so sorry I've been away, been super busy these days :-( but yea I think your last idea is a good one before we figure this one out for good. What do you think?

Kemdole:
Hi @Streppel. Actually, without the lock feature of the example, I could use the cron for a distributed environment by implementing my own lock and MQ within the task function. So the cron instance that trying to run the duplicated job will run the job function successfully and schedule it as the next interval, but no message will be published to the MQ. So no duplicated business logic will be run. So yeah, we can achieve the same without having this feature inside of this project.

I think we can hide the feature in the doc if most users are using this project for a simple purpose.

tonytony2020:
The distributed lock is a very important feature , any updates on this?

Streppel:
@tonytony2020 not on my part. I'm a little busy atm with some personal projects unfortunately. How do you feel about Kemdole's solution of using a lock within the job?

Streppel:
closing this as we removed this broken feature for now

PR #77, opened 2020-11-14
removing lock structure

Streppel:
### What does this do?
Removes the currently non-working distributed locking feature in order to keep the project slimmer and more focused at achieving  v1.0

### Notes
Even though we might introduce something like this in the future, we should aim for simplicity we using the library - that is, parallel job execution control should be primarily left for the user to handle (ie through some queue messaging platform).


arjunmahishi (approved):
LGTM!

JohnRoesler (commented):
Perhaps leave a note in the readme like the Pr description saying it’s up to caller to handle concurrent jobs

Streppel:
@JohnRoesler good idea

JohnRoesler (approved):

PR #79, opened 2020-11-19
refact: scheduleNextRun now uses duration to schedule next run

Streppel:
### What does this do?
refactors scheduleNextRun to use duration to schedule next run, basically laying the ground to use the time package for scheduling jobs

I was starting to refactor our scheduler to use the time package to schedule next runs instead of iterating over our slice and checking if each job should run and realized that we would need to have the duration between our last run and our next expected run.

The old scheduleNextRun was sort of doing this but assigning it directly to the job instead of returning the duration, so I decided to first refactor this part and make the change scope a little more focused instead of bringing all the refactor at once.

The code and the tests related with it ended up becoming simpler, so that was good! Tell me what you guys think. 

Now, we calculate the duration like this:

Job every 3 seconds last run at 01:00:00. Next run should be lastRun + 3 seconds == 01:00:03 (let's call this value Y). So next run will receive the value of Y.Sub(lastRun), which returns 3 seconds. This may seem redundant work but is needed for other usecases that are non-duration type by nature such as days weeks and months.


go.mod
```diff
@@ -2,11 +2,4 @@ module github.com/go-co-op/gocron
 
 go 1.13
 
-require (
-	github.com/go-redis/redis v6.15.5+incompatible
```
Streppel, line 6: also we didn't do go mod tidy from our Lock removal PR so this happened as well

scheduler.go
```diff
@@ -86,124 +86,148 @@ func (s *Scheduler) ChangeLocation(newLocation *time.Location) {
...
+	return job.lastRun
 }
 
-func (s *Scheduler) rescheduleMonth(j *Job, delta time.Time) time.Time {
-	if j.neverRan() { // calculate days to j.dayOfTheMonth
-		jobDay := time.Date(delta.Year(), delta.Month(), j.dayOfTheMonth, 0, 0, 0, 0, s.loc).Add(j.atTime)
-		daysDifference := int(math.Abs(delta.Sub(jobDay).Hours()) / 24)
-		nextRun := s.roundToMidnight(delta)
-		if jobDay.Before(delta) { // shouldn't run this month; schedule for next interval minus day difference
+func (s *Scheduler) calculateMonths(job Job, lastRun time.Time) time.Duration {
```
Streppel, line 138: the calculate*time* methods can be further simplified however I'm thinking about adding even more tests before moving on with any refactor

job.go
```diff
@@ -6,10 +6,12 @@ import (
 	"time"
 )
 
+type jobInterval uint64
+
 // Job struct stores the information necessary to run a Job
 type Job struct {
-	sync.RWMutex
-	interval          uint64                   // pause interval * unit between runs
+	*sync.RWMutex
```
JohnRoesler, line 13: Why pointer to the mutex?

JohnRoesler (approved):
Just one question, otherwise, awesome!

scheduler.go
```diff
@@ -86,124 +86,148 @@ func (s *Scheduler) ChangeLocation(newLocation *time.Location) {
...
+	switch job.unit {
 	case seconds, minutes, hours:
-		j.nextRun = s.rescheduleDuration(j, delta)
+		duration = s.calculateDuration(job)
 	case days:
-		j.nextRun = s.rescheduleDay(j, delta)
+		duration = s.calculateDays(job, lastRun)
 	case weeks:
-		j.nextRun = s.rescheduleWeek(j, delta)
+		{
```
arjunmahishi, line 118: nit: Do we need the curly brace here? I think Go allows us to have multi-line block of code within a `case` without having to enclose it in curly braces.

arjunmahishi (approved):
Awesome!!

scheduler.go
Streppel: nice catch, fixing!

job.go
Streppel: Oh, during development I've switched the Job struct to contain a sync.Locker, and the RWMutex Lock and Unlock functions have pointer receivers, so I had to do this. Then I've switched back to using a RWMutex and forgot to turn it back to value. I guess this is not a big deal (or am I missing something?) but I'll revert it nonetheless!

JohnRoesler (approved):

Streppel:
A little curiosity: I've just calculated the cyclomatic complexity of this repository before and after this PR, and we've lowered our average cyclomatic complexity score from 1.73 to 1.67! :) so that's an improvement

PR #83, opened 2020-11-26
Expose the Job's lastRun, nextRun and runCount 

arjunmahishi:
### What does this do?
Add 3 exported methods to the Job struct to expose lastRun, nextRun and runCount

Yes

- [ ] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -126,15 +126,24 @@ func (j *Job) shouldRun() bool {
 
 // LastRun returns the time the job was run last
 func (j *Job) LastRun() time.Time {
-	return j.lastRun
+	j.Lock()
+	defer j.Unlock()
+	lastRun := j.lastRun
+	return lastRun
 }
 
 // NextRun returns the time the job will run next
 func (j *Job) NextRun() time.Time {
-	return j.nextRun
+	j.Lock()
+	defer j.Unlock()
+	nextRun := j.nextRun
+	return nextRun
 }
 
 // RunCount returns the number of time the job ran so far
 func (j *Job) RunCount() int {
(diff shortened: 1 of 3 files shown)


job.go
```diff
@@ -122,3 +122,18 @@ func (j *Job) LimitRunsTo(n int) {
 func (j *Job) shouldRun() bool {
 	return !j.runConfig.finiteRuns || j.runCount < j.runConfig.maxRuns
 }
+
+// LastRun returns the time the job was run last
+func (j *Job) LastRun() time.Time {
+	return j.lastRun
```
JohnRoesler, line 128: we probably want to lock the mutex here so we don't get a data race if another go routine is currently modifying the job when this is called. You can check the tests with `go test -race github.com/go-co-op/gocron`

Streppel (dismissed):

arjunmahishi:
Thanks for the review guys!  Will close this on the weekend. A little busy today and tomorrow.

job.go
arjunmahishi: This might be a dumb question. But I just saw that we are locking mutex while modifying the `lastRun` value. Do we also need to lock when we read it?

arjunmahishi: This is not showing up in `go test -race`.

JohnRoesler: yes - otherwise you couldn't guarantee the value isn't changing at the moment you read it. This has some good examples https://golangbot.com/mutex/

arjunmahishi: Thanks!

JohnRoesler (approved):

PR #87, opened 2020-12-09
fix runnableJobs where subsequent jobs are never checked

### LINKED ISSUE #85: [FEATURE] - Remove Job from Scheduler Jobs when the Job is supposed to be launch only once.

I've have a Job that i would like to run only once at a specific date

```
job, err = scheduler.Every(1).StartAt(date).SetTag(tag).Do(task)
...
job.LimitRunsTo(1)

```
However when the Job has run i still have a reference to it in the scheduler `scheduler.Jobs()`
 
I have to manually remove this specific Job in my task when the Job is done.

`scheduler.RemoveJobByTag(tag)`

Because if i don't, and if i run a second Job after this, then the second job will never be launched.

When we choose to limit the job to be runned only once `( job.LimitRunsTo(1) )` and its execution has been done then remove this Job from the scheduler Jobs.

JohnRoesler:
### What does this do?
- fixes a bug identified where subsequent jobs are not scheduled if the first job is evaluating false. The `break` in the for loop causes the 2nd element to never be reached. It should be `continue`
- also some spelling fixes

resolves #85 

👍 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -250,8 +250,6 @@ func (s *Scheduler) runnableJobs() []*Job {
 	for _, job := range s.jobs {
 		if s.shouldRun(job) {
 			runnableJobs = append(runnableJobs, job)
-		} else {
-			continue
 		}
 	}
 	return runnableJobs
--- a/scheduler.go
+++ b/scheduler.go
@@ -251,7 +251,7 @@ func (s *Scheduler) runnableJobs() []*Job {
 		if s.shouldRun(job) {
 			runnableJobs = append(runnableJobs, job)
 		} else {
-			break
+			continue
 		}
 	}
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -251,7 +251,7 @@ func (s *Scheduler) runnableJobs() []*Job {
 		if s.shouldRun(job) {
 			runnableJobs = append(runnableJobs, job)
 		} else {
-			break
+			continue
 		}
```
arjunmahishi, line 255: Nit: Isn't this `else` block redundant now? we could just remove it.

arjunmahishi (approved):

arjunmahishi (approved):

PR #89, opened 2020-12-11
Fix: Handle job creation errors better

### LINKED ISSUE #84: [BUG] - Reflect panic with too many jobs possibly
### Describe the bug

reported by @Algalish

```
panic: reflect: call of reflect.Value.Type on zero Value

goroutine 44934 [running]:
reflect.Value.Type(0x0, 0x0, 0x0, 0x0, 0x175fdd5aee8)
        /usr/local/go/src/reflect/value.go:1877 +0x166
github.com/go-co-op/gocron.callJobFuncWithParams(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x175fe0375a8, 0x40fd526000000000, 0x0, 0x0)
       /projects/src/src/github.com/go-co-op/gocron/gocron.go:53 +0xc5
github.com/go-co-op/gocron.(*Job).run(0xc0000dc540)
        /projects/src/src/github.com/go-co-op/gocron/job.go:52 +0x13b
created by github.com/go-co-op/gocron.(*Scheduler).run
        /projects/src/src/github.com/go-co-op/gocron/scheduler.go:290 +0x3f
```

### To Reproduce

```
cronScheduler = gocron.NewScheduler(time.UTC)
	for h := 0; h < 24; h++ {
		for m := 0; m < 60; m += 20 {
			timeString := fmt.Sprint(h) + ":" + fmt.Sprint(m)
			cronScheduler.Every(1).Day().At(timeString).Do(getModDetails)
			log.Println(timeString)
		}
	}
```

arjunmahishi:
### What does this do?
Remove the placeholder job from the scheduler while returning any error from the `Do` function

Closes #84 

Yes

### DIFF
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -846,3 +846,19 @@ func TestRunJobsWithLimit(t *testing.T) {
 	assert.Exactly(t, 1, j1Counter)
 	assert.Exactly(t, 1, j2Counter)
 }
+
+func TestErrorDuringJobCreation(t *testing.T) {
+	// error due to bad time format
+	badTime := "0:0"
+	s := NewScheduler(time.Local)
+	s.Every(1).Day().At(badTime).Do(func() {})
+	assert.Zero(t, len(s.jobs), "The job should be deleted if the time format is wrong")
+
+	// error due to the arg passed to Do() not being a function
+	s.Every(1).Second().Do(1)
+	assert.Zero(t, len(s.jobs), "The job should be deleted if the arg passed to Do() is not a function")
+
+	// positive case
+	s.Every(1).Day().Do(func() {})
+	assert.Equal(t, 1, len(s.jobs))
+}
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -135,6 +135,13 @@ func TestAtFuture(t *testing.T) {
 	nextRun = dayJob.ScheduledTime()
 	assert.Equal(t, expectedStartTime, nextRun)
 	assert.Equal(t, false, shouldBeFalse, "Day job was not expected to run as it was in the future")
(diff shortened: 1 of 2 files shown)


scheduler_test.go
```diff
@@ -846,3 +846,19 @@ func TestRunJobsWithLimit(t *testing.T) {
 	assert.Exactly(t, 1, j1Counter)
 	assert.Exactly(t, 1, j2Counter)
 }
+
+func TestErrorDuringJobCreation(t *testing.T) {
+	// error due to bad time format
+	badTime := "0:0"
+	s := NewScheduler(time.Local)
+	s.Every(1).Day().At(badTime).Do(func() {})
```
JohnRoesler, line 854: Could we move this test case into the `At` func test?

scheduler_test.go
```diff
@@ -846,3 +846,19 @@ func TestRunJobsWithLimit(t *testing.T) {
...
+	assert.Zero(t, len(s.jobs), "The job should be deleted if the time format is wrong")
+
+	// error due to the arg passed to Do() not being a function
+	s.Every(1).Second().Do(1)
```
JohnRoesler, line 858: do we have a test for `Do`? I really like mapping tests directly to functions

scheduler_test.go
arjunmahishi: Done

scheduler_test.go
arjunmahishi: Come to think of it, this function (`TestErrorDuringJobCreation`) is actually testing `Do` function. So, I could rename this function `TestDo`.

Streppel (approved):

JohnRoesler (approved):

PR #90, opened 2020-12-14
feat(remove/job): remove a particular job reference from the schedule…

### LINKED ISSUE #85: [FEATURE] - Remove Job from Scheduler Jobs when the Job is supposed to be launch only once.

dlaweb:
…r after the last execution of the job

### What does this do?

With the scheduler we have the possibility to run a Job only once but when the Job has finished its execution we still have a reference in the Scheduler while it will not be used anymore this can lead to keep obsoletes scheduler Jobs references.

This PR allow the user to add a flag `RemoveAfterLastExec` in order to remove the job after the last job execution.

resolves #85

Yes

- [x] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -31,6 +31,7 @@ type Job struct {
 type runConfig struct {
 	finiteRuns bool
 	maxRuns    int
+	removeAfterLastRun   bool
 }
 
 // NewJob creates a new Job with the provided interval
@@ -146,4 +147,11 @@ func (j *Job) RunCount() int {
 	defer j.Unlock()
 	runCount := j.runCount
 	return runCount
+}
+// RemoveAfterLastRun update the job in order to remove the job after its last exec
+func (j *Job) RemoveAfterLastRun() *Job {
+	j.Lock()
+	defer j.Unlock()
+	j.runConfig.removeAfterLastRun = true
+	return j
 }
--- a/scheduler.go
+++ b/scheduler.go
@@ -456,6 +456,12 @@ func (s *Scheduler) StartImmediately() *Scheduler {
 
 // shouldRun returns true if the Job should be run now
 func (s *Scheduler) shouldRun(j *Job) bool {
+
+	// option remove the job's in the scheduler after its last execution
+	if j.runConfig.removeAfterLastRun && (j.runConfig.maxRuns - j.runCount) == 1 {
+		s.RemoveByReference(j)
+	}
+
 	return j.shouldRun() && s.time.Now(s.loc).Unix() >= j.nextRun.Unix()
 }
 
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -881,3 +881,18 @@ func TestDo(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoveAfterExec(t *testing.T) {
+	s := NewScheduler(time.UTC)
+	s.StartAsync()
+
+	job, err := s.Every(1).StartAt(time.Now().Add(1*time.Second)).Do(task, s)
+	require.NoError(t, err)
+
+	job.LimitRunsTo(1)
+	job.RemoveAfterLastRun()
+
+	time.Sleep(2 * time.Second)
+
+	assert.Zero(t, len(s.Jobs()))
+}
--- a/README.md
+++ b/README.md
@@ -76,6 +76,11 @@ func main() {
 
     // Removing Job Based on Tag
     s2.RemoveJobByTag("tag1")
+    
+    // Remove a Job after its last execution
+    j, _ := s2.Every(1).StartAt(time.Now().Add(30*time.Second)).Do(task)
+    j.LimitRunsTo(1)
+    j.RemoveAfterLastRun()
 
     // Do jobs on specific weekday
     s2.Every(1).Monday().Do(task)
--- a/example_test.go
+++ b/example_test.go
@@ -132,3 +132,11 @@ func ExampleJob_RunCount() {
 	}()
 	<-s.StartAsync()
 }
+
+func ExampleJob_RemoveAfterLastRun() {
+	s := gocron.NewScheduler(time.UTC)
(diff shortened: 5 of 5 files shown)


job.go
```diff
@@ -147,3 +148,10 @@ func (j *Job) RunCount() int {
 	runCount := j.runCount
 	return runCount
 }
+// RemoveAfterLastExec update the job in order to remove the job after its last exec
+func (j *Job)RemoveAfterLastExec() *Job {
```
JohnRoesler, line 152: could you change this to 
```suggestion
func (j *Job)RemoveAfterLastRun() *Job {
```

scheduler.go
```diff
@@ -106,6 +106,11 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 
 	durationToNextRun := s.durationToNextRun(job)
 	job.nextRun = job.lastRun.Add(durationToNextRun)
+
+	// option remove the job's in the scheduler after its last execution
+	if job.runConfig.toRemove && (job.runConfig.maxRuns - job.runCount) == 1 {
+			s.RemoveByReference(job)
```
JohnRoesler, line 112: the indentation here is too much 
```suggestion
		s.RemoveByReference(job)
```

scheduler_test.go
```diff
@@ -881,3 +881,29 @@ func TestDo(t *testing.T) {
...
+
+func TestRemoveAfterExec(t *testing.T) {
+	f := func(s *Scheduler) {
+		fmt.Print("i'm a dummy task")
+	}
+
+	s := NewScheduler(time.UTC)
+	s.StartAsync()
+
+	job1, err := s.Every(1).StartAt(time.Now().Add(10*time.Second)).Do(f, s)
```
JohnRoesler, line 893: could you use the `task` funk defined on line 33?

scheduler_test.go
```diff
@@ -881,3 +881,29 @@ func TestDo(t *testing.T) {
...
+	time.Sleep(15 * time.Second)
+
+	var jobExists bool
+	for _, j := range s.jobs {
+		if j == job1 {
+			jobExists = true
+		}
+	}
+
+	assert.False(t, jobExists)
```
JohnRoesler, line 908: since you're only adding the one job here, you could just assert the len of `s.Jobs()` is 0
```suggestion
	assert.Zero(t, len(s.Jobs()))
```

JohnRoesler:
Thanks for the feature PR!

scheduler.go
arjunmahishi: Also, we are already doing this check [here](https://github.com/go-co-op/gocron/blob/5ed8230dfb9bdf0beb0649cf27c69479dcdc9873/scheduler.go#L459). Would it make more sense to do the cleanup there?
@JohnRoesler @Streppel

JohnRoesler: Yes i think that makes more sense. Move this logic inside `shouldRun` - so when it's checked if a job should run, it also just removes it if it should be removed based on the `toRemove` parameter`.

job.go
```diff
@@ -31,6 +31,7 @@ type Job struct {
 type runConfig struct {
 	finiteRuns bool
 	maxRuns    int
+	toRemove   bool
```
JohnRoesler, line 34:
```suggestion
	removeAfterLastRun   bool
```

scheduler_test.go
```diff
@@ -881,3 +881,18 @@ func TestDo(t *testing.T) {
 		})
 	}
 }
+
+func TestRemoveAfterExec(t *testing.T) {
+	s := NewScheduler(time.UTC)
+	s.StartAsync()
+
+	job, err := s.Every(1).StartAt(time.Now().Add(10*time.Second)).Do(task, s)
```
JohnRoesler, line 889: let's set this to 1 sec so the test doesn't take as long
```suggestion
	job, err := s.Every(1).StartAt(time.Now().Add(1*time.Second)).Do(task, s)
```

scheduler_test.go
```diff
@@ -881,3 +881,18 @@ func TestDo(t *testing.T) {
...
+	require.NoError(t, err)
+
+	job.LimitRunsTo(1)
+	job.RemoveAfterLastRun()
+
+	time.Sleep(15 * time.Second)
```
JohnRoesler, line 895:
```suggestion
	time.Sleep(2 * time.Second)
```

JohnRoesler (approved):

JohnRoesler (approved):

JohnRoesler:
Thank you!

apremalal:
Hi, Thanks for adding this feature. Is it possible to create a release tag with this change?

JohnRoesler:
@apremalal yes we are working on a release soon. This feature may change slightly before the release, but the functionality will be available.

JohnRoesler:
@Streppel you had mentioned why not just remove after the run limit is reached. I think having this available as is with a specific flag is valuable, because a user may not be default want to remove the job. For example, start a job with a limit of 1 run, then later edit the run limit to 2, and then later 3, etc. With this flag, we allow specifically asking to remove the job after the limit is reached.

Streppel:
@JohnRoesler I see... Let's keep it that way, then; I think this makes sense. Even though, creating a new job would be a simpler way of achieving the same result, right?

PR #99, opened 2020-12-27
refact: schedules jobs with time.AfterFunc()

Streppel:
### What does this do?
Main change: scheduler now uses `time.AfterFunc` from the standard ```time``` package to schedule jobs instead of manually sorting and looping through all jobs (yes, finally!)

This brings us:

- assertion that our "run jobs" algorithm works without bugs as we're leaving this logic to a standard package
- greater flexibility in what fields we want to keep in jobs - no need to actually keep other fields like `nextRun` or even our job slice for our run logic (they may still be useful for logging or getting meta info about a job or the scheduler, so they're not removed here)
- minor performance gains as we don't sort and loop through our job array every second anymore
- increased cohesion of our scheduler package 

Unfortunately, after doing this refactor and separating the logic correctly, some tests broke, which made this pr a bit larger than what it needed to be. Fortunately, these failures actually brought minor errors to attention; I fixed them in here, so here's a list of what's different:

- calling `.Close()` without listening to the channel returned by `.StartAsync()` would freeze the application;  fixed by making `stopChan` a buffered channel: now the application doesn't freeze up anymore 
- before, the scheduler invoked each job by calling `go job.Run()`. This method not only calls the job function but also updates the run counter, which fails to remove jobs that start immediately (this passed our tests because all tests used `.At()`). This has been fixed, and now we first update our run counter and only later call the job's function; thus it's launch is actually the last thing to be done in the execution chain 
- refactored `TestExecutionSecond` to `TestImmediateExecution`, as this is what was actually being tested here
- refactored `TestAtFuture` to use `t.Run()` and the new `s.start()` method
- fixed `TestRunJobsWithLimit` to tests cases with immediate execution

None

Yes

### Notes

I did my tests manually here, but if possible guys, take a look at some running examples to also check if they are working correctly. 

A simple example of manual testing function that can be called from any testing file: 
```go
func TestManualTesting(t *testing.T) {
	s := NewScheduler(time.UTC)
	s.Every(1).Second().Do(func() {fmt.Println("job 1: ", time.Now().UTC().Format("15:04:05"))})
	s.Every(5).Second().Do(func() {fmt.Println("job 2: ", time.Now().UTC().Format("15:04:05"))})
	s.StartAsync()
	time.Sleep(11 * time.Second)
}```

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -309,9 +309,10 @@ func (s *Scheduler) Every(interval uint64) *Scheduler {
 }
 
 func (s *Scheduler) run(job *Job) {
-	if !s.running {
+	if !s.IsRunning() {
 		return
 	}
+
 	job.setLastRun(s.now())
 	job.run()
 }
--- a/scheduler.go
+++ b/scheduler.go
@@ -145,10 +145,10 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 
 	durationToNextRun := s.durationToNextRun(lastRun, job)
 	job.setNextRun(lastRun.Add(durationToNextRun))
-	job.timer = time.AfterFunc(durationToNextRun, func() {
+	job.setTimer(time.AfterFunc(durationToNextRun, func() {
 		s.run(job)
 		s.scheduleNextRun(job)
-	})
+	}))
 }
 
 func (s *Scheduler) durationToNextRun(t time.Time, job *Job) time.Duration {
--- a/scheduler.go
+++ b/scheduler.go
@@ -133,7 +133,7 @@ func (s *Scheduler) Location() *time.Location {
 // scheduleNextRun Compute the instant when this Job should run next
 func (s *Scheduler) scheduleNextRun(job *Job) {
 	now := s.now()
-	lastRun := job.lastRun
+	lastRun := job.LastRun()
 
 	// job can be scheduled with .StartAt()
 	if job.neverRan() {
--- a/scheduler.go
+++ b/scheduler.go
@@ -36,28 +36,33 @@ func NewScheduler(loc *time.Location) *Scheduler {
 	}
 }
 
-// StartBlocking starts all the pending jobs using a second-long ticker and blocks the current thread
+// StartBlocking starts all jobs and blocks the current thread
 func (s *Scheduler) StartBlocking() {
 	<-s.StartAsync()
 }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: scheduler.go [hunk 1]
```diff
@@ -46,24 +46,30 @@ func (s *Scheduler) StartAsync() chan struct{} {
...
-				s.setRunning(false)
-				return
+func (s *Scheduler) runJobs() {
+	for _, j := range s.Jobs() {
+		if j.getStartsImmediately() {
+			s.run(j)
+			j.setStartsImmediately(false)
+		}
+		if !j.shouldRun() {
+			if j.getRemoveAfterLastRun() { // TODO: this method seems unnecessary as we could always remove after the run cout has expired. Maybe remove this in the future?
```
JohnRoesler, line 66: 👍

scheduler.go
```diff
@@ -126,41 +132,40 @@ func (s *Scheduler) Location() *time.Location {
 
 // scheduleNextRun Compute the instant when this Job should run next
 func (s *Scheduler) scheduleNextRun(job *Job) {
-	now := s.time.Now(s.Location())
+	now := s.now()
+	lastRun := job.lastRun
```
JohnRoesler, line 136: should do a read lock around the `job.lastRun` ref

JohnRoesler (commented):
looking awesome! just some races to clean up

scheduler.go
JohnRoesler:
```suggestion
	lastRun := job.LastRun()
```

JohnRoesler (approved):

Streppel:
@JohnRoesler thank you so much for the feedback and cleaning up the data race issues!

scheduler.go
```diff
@@ -310,25 +303,12 @@ func (s *Scheduler) Every(interval uint64) *Scheduler {
 	return s
 }
 
-// RunPending runs all the Jobs that are scheduled to run.
-func (s *Scheduler) RunPending() {
-	for _, job := range s.runnableJobs() {
-		s.runAndReschedule(job) // we should handle this error somehow
+func (s *Scheduler) run(job *Job) {
+	if !s.running {
```
Streppel, line 307: @JohnRoesler also, do you think this might help with that issue that we talked earlier on slack?

JohnRoesler (approved):

Issue #102, opened 2021-01-14
[BUG] - StartAt Task doesn't run since 0.5.0


dlaweb:
### Describe the bug

Something is broken since the last release 0.5.0; tasks aren't executed anymore in that specific case :

### To Reproduce

```
package main

import(
	"fmt"
	"time"
	"github.com/go-co-op/gocron"
)

func task() {
	fmt.Println("Im a dummy task")
}
func main() {
	// defines a new scheduler that schedules and runs jobs
	scheduler := gocron.NewScheduler(time.UTC)

	j, _ := scheduler.Every(1).StartAt(time.Now().Add(5*time.Second)).Do(task)
	j.LimitRunsTo(1)
	scheduler.StartAsync()
	
	time.Sleep(15*time.Second)

}

```
### Version

0.5.0

JohnRoesler:
duplicate https://github.com/go-co-op/gocron/issues/98

JohnRoesler:
Check out [v0.5.1](https://github.com/go-co-op/gocron/releases/tag/v0.5.1)

Issue #98, opened 2020-12-25
[BUG] - schedule job at specific time


PolyglotOpenstreetmap:
### Describe the bug

I start a scheduler in my local timezone. Then I read the scheduled jobs from a struct in settings.
I receive the start time as a json timestamp, which is converted to a Golang Time. In the printed output, it seems like this conversion works the way it should.

Also when I ask when the next scheduled job is going to run, I see the correct time.

Alas, when the time comes to run the job, nothing happens.

If I add a recurring job to the queue, say every 2 seconds, I do see the function getting executed with the parameters that were passed.

I also see the tag, I added to the job.

Any idea what I might be doing wrong?

Jo

### To Reproduce

`

func main() {
	location, _ := time.LoadLocation("Europe/Brussels")
	scheduler := gocron.NewScheduler(location)

	for _, task := range settings.Schedule {
		PrettyPrint(task)
		now := nowFormatted()
		if task.StartDate > now {
			var DateRE = regexp.MustCompile(`/Date\((?P<UnixTimestamp>\d+)(?P<TZOffset>[-\+]\d+)\)`)
			tsString := DateRE.FindStringSubmatch(task.StartDate)[1]
			timestamp, _ := strconv.ParseInt(tsString, 10, 64)
			startTime := time.Unix(timestamp/1000, 0)
			job, err := scheduler.Every(1).Day().StartAt(startTime).SetTag([]string{task.ID}).Do(taskWithParams, task)
			// scheduler.Every(2).Seconds().Do(taskWithParams, task)
			PrettyPrint(tsString)
			PrettyPrint(timestamp)
			PrettyPrint(startTime.Format("Mon Jan 2 15:04:05 -0700 MST 2006"))
			_, tt := scheduler.NextRun()
			log.Warn("Next run")
			PrettyPrint(tt.Format("Mon Jan 2 15:04:05 -0700 MST 2006"))
			if err != nil {
				log.Error(err)
			}
			if task.RecurrencePattern == "" {
				job.LimitRunsTo(1)
				job.RemoveAfterLastRun()
			}
	scheduler.StartAsync()
}

`
Output:

`
{
  "Action": 0,
  "EndDate": "/Date(1608919200000+0100)/",
  "Filename": "/home/pi/doSomethingImportant.sh",
  "Id": "8a102123-a91d-474b-8f83-547c9cdcf3ad",
  "Parameters": "",
  "ProcedureName": "main",
  "RecurrencePattern": "",
[...]

PR #103, opened 2021-01-15
fix bug with StartAt() in v0.5.0

JohnRoesler:
### What does this do?
- fixes a bug we introduced in v0.5.0
- introduced `startAtTime` field on the job struct so we know when a job is scheduled to start at a specific time and we should go ahead and start the timer! 

#102 #98 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -142,9 +142,6 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 
 	// job can be scheduled with .StartAt()
 	if job.neverRan() {
-		if !job.NextRun().IsZero() {
-			return // scheduled for future run and should skip scheduling
-		}
 		lastRun = now
 	}
 
@@ -157,6 +154,11 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 }
 
 func (s *Scheduler) durationToNextRun(t time.Time, job *Job) time.Duration {
+	// job can be scheduled with .StartAt() and already has a next run
+	if job.nextRun.After(s.time.Now(s.Location())) {
+		return job.nextRun.Sub(s.time.Now(s.Location()))
+	}
+
 	var duration time.Duration
 	switch job.unit {
 	case seconds, minutes, hours:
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -157,6 +154,11 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 }
 
 func (s *Scheduler) durationToNextRun(t time.Time, job *Job) time.Duration {
+	// job can be scheduled with .StartAt() and already has a next run
+	if job.nextRun.After(s.time.Now(s.Location())) {
+		return job.nextRun.Sub(s.time.Now(s.Location()))
```
Streppel, line 159:
```suggestion
		return job.nextRun.Sub(s.now())
```

scheduler.go
```diff
@@ -157,6 +154,11 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 }
 
 func (s *Scheduler) durationToNextRun(t time.Time, job *Job) time.Duration {
+	// job can be scheduled with .StartAt() and already has a next run
+	if job.nextRun.After(s.time.Now(s.Location())) {
```
Streppel, line 158:
```suggestion
	if job.nextRun.After(s.now()) {
```

Streppel, line 158:
```suggestion
	if job.nextRun.After(t) {
```

Being ```t``` our lastRun time (or now if the job never ran) maybe that'd be the better choice to avoid minimum milliseconds differences between each call

Streppel (approved):

dlaweb:
Still have some issue with the 0.5.1
Infinite loop of task with this code:

```
func task() {
	fmt.Println("Im a dummy task")
	return
}

func main() {
	s1 := gocron.NewScheduler(time.UTC)

	// scheduler starts running jobs and current thread continues to execute
	s1.StartAsync()

	// Remove a Job after its last execution
	j, _ := s1.Every(1).StartAt(time.Now().Add(5 * time.Second)).Do(task)
	j.LimitRunsTo(1)
	time.Sleep(15 * time.Second)
}
```

PR #104, opened 2021-01-22
fix calculateMonths

sdw2330976:
when we use every month at some time,then calculate next run time use base time is not consistency

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -199,7 +199,7 @@ func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) time.Duration {
 				nextRun = nextRun.AddDate(0, int(job.interval), daysDifference)
 			}
 		}
-		return s.until(lastRunRoundedMidnight, nextRun)
+		return s.until(lastRun, nextRun)
 	}
 	nextRun := lastRunRoundedMidnight.Add(job.getAtTime()).AddDate(0, int(job.interval), 0)
 	return s.until(lastRunRoundedMidnight, nextRun)
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -990,11 +990,17 @@ func TestRemoveAfterExec(t *testing.T) {
 }
 
 func TestCalculateMonths(t *testing.T) {
-	s := NewScheduler(time.Local)
+	ft := fakeTime{onNow: func(l *time.Location) time.Time {
+		return time.Date(1970, 1, 1, 12, 0, 0, 0, l)
+	}}
+	s := NewScheduler(time.UTC)
+	s.time = ft
 	s.StartAsync()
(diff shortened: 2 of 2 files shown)


scheduler.go
```diff
@@ -199,7 +199,7 @@ func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) time.Duration {
 				nextRun = nextRun.AddDate(0, int(job.interval), daysDifference)
 			}
 		}
-		return s.until(lastRunRoundedMidnight, nextRun)
+		return s.until(lastRun, nextRun)
```
arjunmahishi, line 202: Could you please include a test that fails in the case of  `lastRunRoundedMidnight` and passes in the case of `lastRun`? That would be super helpful.

arjunmahishi (commented):
Hey @sdw2330976, thanks for your contribution! 🎉

JohnRoesler (approved):
lgtm

scheduler_test.go
```diff
@@ -988,3 +988,13 @@ func TestRemoveAfterExec(t *testing.T) {
...
+
+func TestCalculateMonths(t *testing.T) {
+	s := NewScheduler(time.Local)
+	s.StartAsync()
+	job, _ := s.Every(1).Month(1).At("10:00").Do(func() {
+		fmt.Println("hello task")
+	})
+	fmt.Println(job.ScheduledTime().Format("2006-01-02 15:04:05"))
+	time.Sleep(2 * time.Second)
+}
```
JohnRoesler, line 1000: we can make this test not concerned with the current time and fake the time for consistent results
```suggestion
func TestCalculateMonths(t *testing.T) {
	ft := fakeTime{onNow: func(l *time.Location) time.Time {
		return time.Date(1970, 1, 1, 12, 0, 0, 0, l)
	}}
	s := NewScheduler(time.UTC)
	s.time = ft
	s.StartAsync()
	job, err := s.Every(1).Month(1).At("10:00").Do(func() {
		fmt.Println("hello task")
	})
	require.NoError(t, err)
	s.Stop()

	assert.Equal(t, s.time.Now(s.location).AddDate(0, 1, 0).Month(), job.nextRun.Month())
}
```

scheduler.go
```diff
@@ -191,7 +191,7 @@ func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) time.Duration {
...
```
JohnRoesler, line 194: @sdw2330976 do you have a time scenario where this fails with `lastRunRoundedMidnight`? We can fake the time for a test, but ideally we could prove the old logic doesn't work with our test

JohnRoesler (approved):

PR #105, opened 2021-01-23
Remove stopChan from Scheduler 

### LINKED ISSUE #100: [FEATURE] - Allow scheduler to be stopped with Stop()

@Streppel 

> We should stop returning it and make the scheduler stoppable by calling .Stop(), but that's another story

arjunmahishi:
### What does this do?
Remove stopChan from Scheduler to restrict stopping the scheduler only to the Stop function

Closes #100 

The `StartAsync()` function will not return the `stopChan` anymore. This is a backward incompatible change

- [ ] Updated `README.md`

### DIFF
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -78,9 +78,9 @@ func TestExecutionSeconds(t *testing.T) {
 		}
 	})
 
-	stop := sched.StartAsync()
+	sched.StartAsync()
 	<-jobDone // Wait job done
-	close(stop)
+	sched.Stop()
 
 	mu.RLock()
 	defer mu.RUnlock()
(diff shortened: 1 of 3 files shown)


scheduler_test.go
```diff
@@ -389,9 +389,9 @@ func TestScheduler_Stop(t *testing.T) {
 	})
 	t.Run("stops a running scheduler through StartAsync chan", func(t *testing.T) {
```
Streppel, line 390:
```suggestion
	t.Run("stops a running scheduler calling .Stop()", func(t *testing.T) {
```

Streppel (dismissed):
LGTM!

JohnRoesler (approved):

Issue #107, opened 2021-01-25
[BUG] - RemoveJob does not work in v0.5.1


huage1994:
### Describe the bug

Method RemoveJobByTag does not work in v0.5.1 ,  it can work in v0.4.0

### To Reproduce

run code in v0.4.0

```
package main

import (
	"fmt"
	"github.com/go-co-op/gocron"
	"time"
)

func main() {
	s1 := gocron.NewScheduler(time.UTC)
	tag1 := []string{"tag1"}
	tag2 := []string{"tag2"}
	s1.Every(3).Seconds().SetTag(tag1).Do(func() {
		fmt.Println(len(s1.Jobs()), "I am job 1", time.Now())
	})

	jobz, _ := s1.Every(3).Seconds().SetTag(tag2).Do(func() {
		fmt.Println(len(s1.Jobs()), "I am job 2", time.Now())
	})
	print(jobz)
	s1.StartAsync()

	time.Sleep(3 * time.Second)

	s1.RemoveJobByTag("tag2")

	fmt.Println(s1.Jobs())
	print("------------------------")

	time.Sleep(100 * time.Second)

}
```

Streppel:
@huage1994 thanks for reporting it! Will take a look.

Nikkhah:
Hi
I have this problem too. Could you find the solution?

huage1994:
> Hi
> I have this problem too. Could you find the solution?

Just now upload fixed code to my forked repo.

Streppel:
Sorry guys I'm short on time, as soon as I get my eyes on this I'll let you know here. Probably later today

JohnRoesler:
This is released in [v0.6.0](https://github.com/go-co-op/gocron/releases/tag/v0.6.0)!

PR #110, opened 2021-01-31
fix [bug](https://github.com/go-co-op/gocron/issues/107) when removin…

huage1994:
…g jobs

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -342,6 +342,8 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
 	for _, job := range s.Jobs() {
 		if !shouldRemove(job) {
 			retainedJobs = append(retainedJobs, job)
+		} else {
+			job.timer.Stop()
 		}
 	}
 	s.setJobs(retainedJobs)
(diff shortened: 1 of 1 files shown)


JohnRoesler:
I'm on mobile util tomorrow so I can't look into the test failure yet - but looks like we're getting a nil pointer reference now.

scheduler.go
```diff
@@ -342,6 +342,8 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
 	for _, job := range s.Jobs() {
 		if !shouldRemove(job) {
 			retainedJobs = append(retainedJobs, job)
+		} else {
+			job.timer.Stop()
```
JohnRoesler, line 346: the timer may be nil here if the job has never been started. I would suggest adding a function to the job file like:

```
func (j *Job) stopTimer() {
	if j.timer != nil {
		j.timer.Stop()
	}
}
```

and the call that function instead of `job.timer.Stop()` as `job.stopTimer()`

JohnRoesler:
I pulled your commit into https://github.com/go-co-op/gocron/pull/112

PR #113, opened 2021-02-03
Allow starting Jobs in the past (#106)

### LINKED ISSUE #106: [BUG] - `StartAt` in the past causes next execution time to calculate based on `Now`
### Using `StartAt` with a date in the past doesn't behave as expected

If I am creating a scheduled job and use `StartAt` with a date in the past, I expected the next execution time to be calculated based on that start date and the defined interval, but instead it uses the current time and the interval.

**I am not sure if that is a bug or intended behavior** so I chose to open this issue before creating a PR. If you support this change, I can go ahead and do the PR myself.

### To Reproduce

```go
package main

import (
	"fmt"
	"time"

	"github.com/go-co-op/gocron"
)

func task() {
	fmt.Println("EXECUTING")
}

func main() {
	scheduler := gocron.NewScheduler(time.Local)

	job, _ := scheduler.
		Every(24).
		Hours().
		StartAt(time.Now().Add(-24 * time.Hour).Add(10 * Minute)).
		Do(task)

	scheduler.StartAsync()
	fmt.Println(job.NextRun())
}
```

### Version

`v0.5.1`

### Expected behavior

The expected output of the code above would be today, at the current time + 10 minutes. Instead, it is 24 hours from the current time which does not fit in the configured schedule.

### Additional context

My use-case here is an [automated garden](https://github.com/calvinmclean/automated-garden) where I have a `Plant` struct containing a `StartDate` to record when it was first planted and serve as the starting point for my watering interval. I want to be able to restart my application, read the `StartDate` from a database or configuration file and run the scheduled jobs based on that instead of based on when the program started.

I implemented a workaround in my code but wanted to have the opportunity to include it in `goCron` if that is a desirable feature/fix.

calvinmclean:
### What does this do?

This PR adds a loop to increment the Job's `startAtTime` until it is in the future. It will only need to do this if a Job is intentionally set to start in the past, not if it is just defaulted to 0.

Resolves #106 

- If someone is intentionally scheduling something to start at 0 time, it might not work as expected.

Yes

Yes

- [ ] Updated `README.md`

### Notes
See #106 for more information

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -133,6 +133,10 @@ func (s *Scheduler) scheduleNextRun(job *Job) {
 	lastRun := job.LastRun()
 
 	if job.neverRan() {
+		// Increment startAtTime until it is in the future
+		for job.startAtTime.Before(now) && !job.startAtTime.IsZero() {
+			job.startAtTime = job.startAtTime.Add(s.durationToNextRun(job.startAtTime, job))
+		}
 		lastRun = now
 	}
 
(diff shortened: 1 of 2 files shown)


JohnRoesler (approved):
thanks for the contribution!

PR #116, opened 2021-02-04
add job limit and remove after last run to chain

### LINKED ISSUE #115: [BUG] - Job with LimitRunsTo(1) runs twice
### Describe the bug

If you add a job to **a running scheduler** and set its **run limits to 1**, then the job is executed **one extra time**. It runs 2 times instead of 1 time.

### To Reproduce

1. StartAsync a scheduler 
2. Add a new job to the scheduler
3. Set the new job's run limit to 1

```go

s.StartAsync()

var job *gocron.Job
job, _ = s.Every(1).Second().Do(func() {
  fmt.Println(job.LastRun())
  fmt.Println(job.RunCount())
  fmt.Println()
})

job.LimitRunsTo(1)

time.Sleep(time.Second * 10)
```

The program output:
```
2021-02-04 07:53:52.5575732 +0000 UTC
1

2021-02-04 07:53:53.5576467 +0000 UTC
2
```

### Version

Current master branch. 7601804f2dea4e0e55d415abfb211f8ec68edf14

### Expected behavior

The job is executed exectly once.

### Additional context

Well, I did not investigate deeply, but I guess that is just a mistake in code, because other limits (more than 1) work correctly.

Correct code:
```go

s.StartAsync()

var job *gocron.Job
job, _ = s.Every(1).Second().Do(func() {
  fmt.Println(job.LastRun())
  fmt.Println(job.RunCount())
  fmt.Println()
})

job.LimitRunsTo(2) // now set to 2

time.Sleep(time.Second * 10)
```

Output as expected with 2 runs:
```
2021-02-04 08:03:13.5170117 +0000 UTC
1

2021-02-04 08:03:14.5185231 +0000 UTC
2
```
But I also see a different problem with `LimitRunsTo`.

I guess there's a problem with concurrency.

I set job's run limits after the job is added to scheduler. That means that the job may be scheduled an unlimited number of times (theoratically) before the `LimitRunsTo` is executed.

I suggest adding `LimitRunsTo` analog to the new job creating pipeline. Like this:
```go
s.Every(1).Second().LimitRuns(2).Do(func() {})
```

JohnRoesler:
### What does this do?

- adds `LimitRunsTo` and `RemoveAfterLastRun` to scheduler api to allow changing a behavior where a job added to a running scheduler cannot be limited to a single run.

closes #115 

👍  and cleaned up a bunch

- [ ] Updated `README.md`


job.go
```diff
@@ -175,9 +175,12 @@ func (j *Job) Weekday() (time.Weekday, error) {
...
 
-// LimitRunsTo limits the number of executions of this
-// job to n. However, the job will still remain in the
-// scheduler
+// LimitRunsTo limits the number of executions of this job to n. However,
+// the job will still remain in the scheduler
+// Note: if a job is added to a running scheduler and this method is used
+// you may see the job run more than the limit as it's scheduled immediately
+// upon being added to the scheduler. It's recommended to use the LimitRunsTo()
+// when scheduling the job scheduler.LimitRunsTo(1).Do()
```
JohnRoesler, line 183:
```suggestion
// LimitRunsTo limits the number of executions of this job to n.
// The job will remain in the scheduler.
// Note: If a job is added to a running scheduler and this method is used
// you may see the job run more than the set limit as job is scheduled immediately
// by default upon being added to the scheduler. It is recommended to use the
// LimitRunsTo() func on the scheduler chain when scheduling the job.
// For example: scheduler.LimitRunsTo(1).Do()
```

job.go
```diff
@@ -187,6 +190,21 @@ func (j *Job) LimitRunsTo(n int) {
...
 
+// RemoveAfterLastRun sets the job to be removed after it's last run (when limited)
+func (j *Job) RemoveAfterLastRun() *Job {
+	j.Lock()
+	defer j.Unlock()
+	j.runConfig.removeAfterLastRun = true
+	return j
+}
+
+// TODO: this method seems unnecessary as we could always remove after the run count has expired. Maybe remove this in the future?
```
JohnRoesler, line 202:
```suggestion
```

PR #119, opened 2021-02-05
Every() accepts int, time.Duration, string to allow any custom duration between runs

### LINKED ISSUE #44: [FEATURE] - Add a way to use a custom time.Duration
Hello :wave: 
I'd like to use a custom `time.Duration`. From what I can see in the code of this lib, the user is forced to set a unit, otherwise this error is returned:

```go
// Error declarations for gocron related errors
var (
	// ...
	ErrPeriodNotSpecified  = errors.New("unspecified job period")
	// ...
)

func (j *Job) setPeriodDuration() error {
	interval := time.Duration(j.interval)

	switch j.unit {
	case seconds:
		j.periodDuration = interval * time.Second
	case minutes:
		j.periodDuration = interval * time.Minute
	case hours:
		j.periodDuration = interval * time.Hour
	case days:
		j.periodDuration = interval * time.Hour * 24
	case weeks:
		j.periodDuration = interval * time.Hour * 24 * 7
	default:
		return ErrPeriodNotSpecified
	}
	return nil
}
```

Simply removing that default case would allow users to do something like that (since `time.Duration`'s underlying type is an `int64`):

```go
j, err := scheduler.Every(uint64(mycustomduration)).Do(...)
```

Another solution would be to add a noop unit that doesn't multiply the value of the given duration, or add a way of directly use a `time.Duration` using another method (which would avoid the `interval := time.Duration(j.interval)` cast), something like `sch.WithDuration` or `sch.EveryDuration`, I'm havint trouble finding a good name to be honest :smile: )

JohnRoesler:
### What does this do?
Changes the `Every()` func in the scheduler allowing users to schedule any time.Duration between job runs via:
- int + unit (e.g. Seconds)
- time.Duration
- string representation of time.Duration e.g. `1s`

closes #44 

👍 

- [ ] Updated `README.md`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -46,6 +46,12 @@ func ExampleScheduler_StartAsync() {
 	s.StartAsync()
 }
 
+func ExampleScheduler_EveryDuration() {
+	s := gocron.NewScheduler(time.UTC)
+	_, _ = s.EveryDuration(500 * time.Millisecond).Do(task)
+	s.StartAsync()
+}
+
 func ExampleScheduler_StartAt() {
 	s := gocron.NewScheduler(time.UTC)
 	specificTime := time.Date(2019, time.November, 10, 15, 0, 0, 0, time.UTC)
--- a/example_test.go
+++ b/example_test.go
@@ -46,9 +46,11 @@ func ExampleScheduler_StartAsync() {
 	s.StartAsync()
 }
 
-func ExampleScheduler_EveryDuration() {
+func ExampleScheduler_Every() {
 	s := gocron.NewScheduler(time.UTC)
-	_, _ = s.EveryDuration(500 * time.Millisecond).Do(task)
+	_, _ = s.Every(1).Second().Do(task)
+	_, _ = s.Every(1 * time.Second).Do(task)
+	_, _ = s.Every("1s").Do(task)
 	s.StartAsync()
 }
 
--- a/gocron.go
+++ b/gocron.go
@@ -45,6 +45,7 @@ const (
 	days
 	weeks
 	months
+	duration
 )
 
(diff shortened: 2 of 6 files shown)


example_test.go
```diff
@@ -46,6 +46,12 @@ func ExampleScheduler_StartAsync() {
 	s.StartAsync()
 }
 
+func ExampleScheduler_EveryDuration() {
+	s := gocron.NewScheduler(time.UTC)
+	_, _ = s.EveryDuration(500 * time.Millisecond).Do(task)
```
arjunmahishi, line 51: Just a thought:
Can we not make this work under `Every()` itself?

Maybe `Every()` can take `interface{}` as the arg instead of `int`. And internally we can check if it's of type `int` or `time.Duration`

Perhaps we can also allows users to give a string duration like "1s", "5m", "3h" etc

JohnRoesler: That's an interesting idea! Yes, I think we could go this route and would be simpler api with only `Every`.

arjunmahishi: Looks like passing just a number will make it an `int` by default. We might have to typecast all the int types to `uint64`
https://play.golang.org/p/7NM3LiwH03J

Streppel (approved):
Amazing @JohnRoesler! Loved this! How about updating our README with the examples from `ExampleScheduler_Every` as well? EDIT: nvm, just saw that we have a checkbox for that too, sorry!

JohnRoesler:
@Streppel what about removing all the read me examples in favor of the go docs?

Streppel:
> @Streppel what about removing all the read me examples in favor of the go docs?

I like this idea! Keep everything in one place. I'm all in for this.

PR #122, opened 2021-02-09
implement executor to reduce go routines / mutexes

### LINKED ISSUE #101: Backend design improvement - Scheduler + Exector
The current design involves a lot of struct accessing across multiple goroutines, resulting in a many mutexes to avoid race conditions.

The proposal here is to move to a new backend design as such:

- Executor: Instead of the scheduler kicking of jobs by calling the `Job.run()` func, the executor will listen on a channel for job functions to execute. This removes the need for multiple goroutines to reference the jobs themselves and we can pass the functions + params across the channel only.
- Scheduler: The scheduler will operate the same as it does, except that it sends on a channel rather than calling `Job.run()` in a goroutine

A start: https://github.com/go-co-op/gocron/compare/master...JohnRoesler:executor?expand=1

JohnRoesler:
### What does this do?

Spins up an executor that runs job functions passed to it on a chan. This changes it so that only the function of the job is passed to a separate goroutine and reduces the references to the job object by different goroutines.

resolves #101 

implicitly tested by all tests

- [x] Updated `README.md`
  - added a link to our examples on the go docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ See also these two great articles:
 
 If you want to chat, you can find us at Slack! [<img src="https://img.shields.io/badge/gophers-gocron-brightgreen?logo=slack">](https://gophers.slack.com/archives/CQ7T0T1FW)
 
+## Examples
+
+Take a look in our [go docs](https://pkg.go.dev/github.com/go-co-op/gocron#pkg-examples)
+
 ## FAQ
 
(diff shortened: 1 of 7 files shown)


Streppel (approved):

Streppel:
This is amazing!!! Thanks @JohnRoesler!

PR #124, opened 2021-02-20
fix race condition

amirhosseinvz94:
### What does this do?

jobs() method lock the jobs to get list of them but after we call Jobs(),this method lock jobs until get list of them, but after this method finished, the caller of this method should need lock the jobs because maybe it need to check current state and no past state, and if in this state removebytag() remove any job this may generate race condition (because used append() function on job in removebytag())

- [ ] Updated `README.md`


JohnRoesler (approved):
Thanks!

PR #126, opened 2021-02-23
Dustin decker/max concurrent jobs

### LINKED ISSUE #53: [FEATURE] - Maximum concurrent jobs

JohnRoesler:
### What does this do?
- Adds the ability to limit the maximum concurrently running jobs on the scheduler
- Adds modes for the max concurrency, of reschedule or wait

resolves #53 
closes #63 

👍 

- [ ] Updated `README.md`


JohnRoesler:
if ya'll wouldn't mind taking a look and let me know what you think of this implementation, I would greatly appreciate it! (i tagged everyone involved in the other open PR)

@dustin-decker, @temamagic, @farwydi, @Streppel, @arjunmahishi

executor.go
```diff
@@ -25,6 +46,26 @@ func (e *executor) start() {
...
+						switch e.limitMode {
+						case RescheduleMode:
+							return
+						case WaitMode:
+							for {
+								if !e.maxRunningJobs.TryAcquire(1) {
+									continue
+								} else {
+									break
+								}
```
Streppel, line 61: wdyt?
```suggestion
								if e.maxRunningJobs.TryAcquire(1) {
									break
								}
```

executor.go
```diff
@@ -25,6 +46,26 @@ func (e *executor) start() {
 			go func() {
 				defer wg.Done()
 
+				if e.maxRunningJobs != nil {
+					if !e.maxRunningJobs.TryAcquire(1) {
+
+						switch e.limitMode {
```
Streppel, line 52: really liked this approach

Streppel (approved):

executor.go
```diff
@@ -25,6 +46,26 @@ func (e *executor) start() {
...
+						case RescheduleMode:
+							return
+						case WaitMode:
+							for {
```
JohnRoesler, line 56: realized I need a way to stop this for loop if the scheduler is stopped - working on it

Streppel: oh, just realized we might need to take into consideration job removals as well right?

PR #130, opened 2021-02-28
wrap errors

JohnRoesler:
### What does this do?
- when multiple errors occur, wrap them so the user can unwrap and check what happened
- collapse/clean up tests

- [ ] Updated `README.md`

### DIFF
--- a/gocron.go
+++ b/gocron.go
@@ -8,6 +8,7 @@ package gocron
 
 import (
 	"errors"
+	"fmt"
 	"reflect"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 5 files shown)


gocron.go
```diff
@@ -8,6 +8,7 @@ package gocron
 
 import (
 	"errors"
+	"fmt"
```
gopherine, line 11: You should not really print anything on console for the package

Streppel: From what I can see the fmt package here is used just to wrap errors, so this should be fine :-)

Issue #142, opened 2021-03-23
[BUG] - job isn't executed silently if wrong list of params passed  


mikhailbolshakov:
### Describe the bug

List of params passed to Do method must correspond to function, otherwise job isn't executed silently

### To Reproduce

The following test reproduces the bug 

``
func Test_SpecificTime_WithIncorrectParams(t *testing.T) {

	quit := make(chan bool)

	action := func(s1, s2 string) {
		fmt.Printf("fired: %s, %s", s1, s2)
		quit <- true
	}

	s := gocron.NewScheduler(time.UTC)
	s.StartAsync()

	j, err := s.Every(1).Day().StartAt(time.Now().UTC().Add(time.Second * 10)).Do(action, "s1")
	if err != nil {
		t.Fatal()
	}
	fmt.Println("next run = ", j.NextRun())

	select {
	case <-time.After(time.Second * 20):
		fmt.Println("time out")
		t.Fatal()
	case <-quit:
		fmt.Println("passed")
	}

}
``

### Version

1.0.0

### Expected behavior

I suggest to return a n error from Do func if the list of passed params is wrong.
I could prepare a PR if you aren't against

mikhailbolshakov:
@JohnRoesler 
Could you pls create a new tag
thanks

JohnRoesler:
https://github.com/go-co-op/gocron/releases/tag/v1.1.0 is out

PR #144, opened 2021-03-24
#142 validation of params is added

mikhailbolshakov:
### What does this do?

It introduces additional validation of params in Scheduler.Do. If list of passed params doesn't correspond to original function params it returns an error.

#142 

TestScheduler_LessParamsThanExpected
TestScheduler_MoreParamsThanExpected

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -532,7 +532,8 @@ func (s *Scheduler) Do(jobFun interface{}, params ...interface{}) (*Job, error) 
 	f := reflect.ValueOf(jobFun)
 	if len(params) != f.Type().NumIn() {
 		s.RemoveByReference(job)
-		return nil, ErrWrongParams
+		job.error = wrapOrError(job.error, ErrWrongParams)
+		return nil, job.error
 	}
 
 	fname := getFunctionName(jobFun)
--- a/scheduler.go
+++ b/scheduler.go
@@ -529,6 +529,12 @@ func (s *Scheduler) Do(jobFun interface{}, params ...interface{}) (*Job, error) 
 		return nil, ErrNotAFunction
 	}
 
+	f := reflect.ValueOf(jobFun)
+	if len(params) != f.Type().NumIn() {
+		s.RemoveByReference(job)
+		return nil, ErrWrongParams
+	}
+
 	fname := getFunctionName(jobFun)
 	job.functions[fname] = jobFun
 	job.params[fname] = params
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -1058,3 +1058,31 @@ func TestScheduler_TagsUnique(t *testing.T) {
 	assert.EqualError(t, err, ErrTagsUnique(bar).Error())
 
 }
+
+func TestScheduler_LessParamsThanExpected(t *testing.T) {
+
+	s := NewScheduler(time.UTC)
+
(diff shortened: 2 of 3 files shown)


JohnRoesler:
@mikhailbolshakov thanks for the PR! this is great stuff. I'm going to drop a couple of comments just on the testing format.

scheduler_test.go
```diff
@@ -1058,3 +1058,31 @@ func TestScheduler_TagsUnique(t *testing.T) {
...
+
+	f := func(s1, s2 string) {
+		fmt.Println("ok")
+	}
+
+	p := []interface{}{"p1", "p2", "p3"}
+	_, err := s.Every(1).Days().StartAt(time.Now().UTC().Add(time.Second*10)).Do(f, p...)
+	assert.EqualError(t, err, ErrWrongParams.Error())
+
+}
```
JohnRoesler, line 1088: could you make the test a table test, and change the name to be for the `Do` func as i have below in the suggestion?

```suggestion
func TestScheduler_DoParameterValidation(t *testing.T) {

	testCases := []struct {
		description string
		parameters  []interface{}
	}{
		{"less than expected", []interface{}{"p1"}},
		{"more than expected", []interface{}{"p1", "p2", "p3"}},
	}
	
	for _, tc := range testCases {
		t.Run(tc.description, func(t *testing.T) {
			s := NewScheduler(time.UTC)

			f := func(s1, s2 string) {
				fmt.Println("ok")
			}
			
			_, err := s.Every(1).Days().StartAt(time.Now().UTC().Add(time.Second*10)).Do(f, tc.parameters...)
			assert.EqualError(t, err, ErrWrongParams.Error())
		})
	}
}
```

mikhailbolshakov: Hey @JohnRoesler 
Thanks for the suggestion, just push it

scheduler.go
```diff
@@ -529,6 +529,12 @@ func (s *Scheduler) Do(jobFun interface{}, params ...interface{}) (*Job, error)
 		return nil, ErrNotAFunction
 	}
 
+	f := reflect.ValueOf(jobFun)
+	if len(params) != f.Type().NumIn() {
+		s.RemoveByReference(job)
+		return nil, ErrWrongParams
```
JohnRoesler, line 535:
```suggestion
		s.RemoveByReference(job)
		return nil, ErrWrongParams
```
```suggestion
		s.RemoveByReference(job)
		job.error = wrapOrError(job.error, ErrWrongParams)
		return nil, job.error
```

JohnRoesler (approved):
thanks for the contribution @mikhailbolshakov

mikhailbolshakov:
@JohnRoesler 
Could you please create a new tag

mikhailbolshakov:
@JohnRoesler 
BTW, could u pls explain your last change where you wrapped the error because it's not quite obvious. As I see the only chance for this wrap to make any sense if s.RemoveByReference(job) populates job.error (otherwise job.error is checked with code above) but I didn't find such logic. 
thanks

JohnRoesler:
Multiple errors can occur in the scheduler chain, for example in the Every func or At, and so we use the error wrapping to return all the errors in job.error so the caller can check what went wrong.

PR #159, opened 2021-04-13
handle occasional occurence of double run on daily job

JohnRoesler:
### What does this do?

we see an occasional instance of a daily job running twice within a second. this seems to be due to the job running too fast and starting a fraction of a second early such that it runs during the 1 second window before it should. this fix handles that case for daily jobs as those only run once at a given even hour/min/sec time.

https://github.com/go-co-op/gocron/issues/52#issuecomment-816502678

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -213,31 +213,31 @@ func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) time.Duration {
 		daysDifference := int(math.Abs(lastRun.Sub(jobDay).Hours()) / 24)
 		nextRun := s.roundToMidnight(lastRun).Add(job.getAtTime())
 		if jobDay.Before(lastRun) { // shouldn't run this month; schedule for next interval minus day difference
-			nextRun = nextRun.AddDate(0, int(job.interval), -daysDifference)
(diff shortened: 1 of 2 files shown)


scheduler.go
```diff
@@ -258,18 +258,26 @@ func (s *Scheduler) calculateTotalDaysDifference(lastRun time.Time, daysToWeekda
 }
 
 func (s *Scheduler) calculateDays(job *Job, lastRun time.Time) time.Duration {
+
 	if job.interval == 1 {
 		lastRunDayPlusJobAtTime := time.Date(lastRun.Year(), lastRun.Month(), lastRun.Day(), 0, 0, 0, 0, s.Location()).Add(job.getAtTime())
+
+		// handle occasional occurrence of job running to quickly / too early such that last run was within a second of now
+		lastRunUnix, nowUnix := job.LastRun().Unix(), s.time.Now(s.location).Unix()
```
Streppel, line 266:
```suggestion
		lastRunUnix, nowUnix := job.LastRun().Unix(), s.now().Unix()
```

Streppel (approved):

Streppel (approved):

PR #160, opened 2021-04-13
clear the job context for singleton on update

### LINKED ISSUE #157: Update() won't work for singleton mode jobs
When a job is running in singleton mode, if you call Update(), the job will never run again.

JohnRoesler:
### What does this do?

singleton jobs as well as waitmode jobs listen for the job context to be cancelled. `Update()` needs to clear that otherwise all subsequent runs are skipped because the context.Done() check is always true.

resolves #157 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -1,6 +1,7 @@
 package gocron
 
 import (
+	"context"
 	"fmt"
 	"math"
 	"reflect"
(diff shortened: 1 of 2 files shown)


Streppel (approved):
:+1:

Issue #152, opened 2021-04-07
s.NotAtStart().Every(1).Hours || s.AwaitFirstSchedule()


bobmcallan:
To configure an efficient service start-up, can you please either add a function to prevent the job from starting when the scheduler starts. Or, a global setting to not start all jobs when the scheduler starts.

`
s.NotAtStart().Every("5m").Do(func(){ ... })

or 

s.AwaitFirstSchedule()

`
Thanks,
Bob

JohnRoesler:
Hello @bobmcallan 👋 

You could currently accomplish this with the [`StartAt()`](https://pkg.go.dev/github.com/go-co-op/gocron#Scheduler.StartAt) method, for example:

```
s := gocron.NewScheduler(time.UTC)	
fiveMinute := time.Minute * 5
s.Every(fiveMinute).StartAt(time.Now().Add(fiveMinute)).Do(func(){ log.Println("hello") })
```

I could see value in the global option you suggest to allow defaulting the scheduler to start the jobs only on their first interval. Thanks for bringing this up.

JohnRoesler:
I have a PR coming soon for this feature

JohnRoesler:
This is in latest release

PR #162, opened 2021-04-15
add WaitForSchedule(s) funcs

JohnRoesler:
### What does this do?
- adds `WaitForSchedule` and `WaitForSchedules` to allow setting per job or for all jobs to wait for the first scheduled run rather than running immediately
- un-exports the job's `NewJob` method. There was no use for it to be exported as we don't accept a newly created job struct in any scheduler methods. alters the signature to handle the wait for interval logic

#152 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -884,11 +884,11 @@ func (s *Scheduler) newJob(interval int) *Job {
 	return newJob(interval, !s.waitForInterval)
 }
 
-// WaitForSchedules defaults the scheduler to create all
+// WaitForScheduleAll defaults the scheduler to create all
 // new jobs with the WaitForSchedule option as true.
 // The jobs will not start immediately but rather will
 // wait until their first scheduled interval.
-func (s *Scheduler) WaitForSchedules() {
+func (s *Scheduler) WaitForScheduleAll() {
 	s.waitForInterval = true
 }
 
--- a/scheduler.go
+++ b/scheduler.go
@@ -881,10 +881,7 @@ func (s *Scheduler) cron(cronExpression string, withSeconds bool) *Scheduler {
 }
 
 func (s *Scheduler) newJob(interval int) *Job {
-	if s.waitForInterval {
-		return newJob(interval, false)
-	}
-	return newJob(interval, true)
+	return newJob(interval, !s.waitForInterval)
 }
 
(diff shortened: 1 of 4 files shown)


scheduler.go
```diff
@@ -878,3 +879,26 @@ func (s *Scheduler) cron(cronExpression string, withSeconds bool) *Scheduler {
 	s.setJobs(append(s.Jobs(), job))
 	return s
 }
+
+func (s *Scheduler) newJob(interval int) *Job {
+	if s.waitForInterval {
+		return newJob(interval, false)
+	}
+	return newJob(interval, true)
```
arjunmahishi, line 887: NIT
```suggestion
	return newJob(interval, !s.waitForInterval)
```

scheduler.go
```diff
@@ -878,3 +879,26 @@ func (s *Scheduler) cron(cronExpression string, withSeconds bool) *Scheduler {
...
+}
+
+// WaitForSchedules defaults the scheduler to create all
+// new jobs with the WaitForSchedule option as true.
+// The jobs will not start immediately but rather will
+// wait until their first scheduled interval.
+func (s *Scheduler) WaitForSchedules() {
```
arjunmahishi, line 894: I think the users might think `WaitForSchedules()` and `WaitForSchedule()` have the same behavior. Because we have other methods like `Second()`-`Seconds()`,  `Minute()`-`Minutes()` etc which in spite of being singular and plural have the same behavior. 

The comments are clear enough, but maybe a different name for `WaitForSchedules` might be nice.

JohnRoesler: `WaitForSchedulesAllJobs()`

arjunmahishi: Sure. Or of the word "Hold" makes sense, they can both be changed to `HoldJobTillFirstSchedule` and `HoldAllJobsTillFirstSchedule`

arjunmahishi: We could also skip the word "First" to make it shorter :P

JohnRoesler: how about `WaitForScheduleAll` i pushed that, let me know

arjunmahishi: Sure 👍🏽

arjunmahishi (approved):

Issue #127, opened 2021-02-25
[FEATURE] - Support running weekly schedule on multiple days of the week


JohnRoesler:
### Is your feature request related to a problem? Please describe

This would be nice to allow a single job to be run multiple days of the week.

### Describe the solution you'd like

```golang
s.Every(1).Week().Monday().Thursday().Saturday().At("02:00").Do(func() {})
```

### Describe alternatives you've considered

```golang
s.Every(1).Week().Monday().At("02:00").Do(func() {})
s.Every(1).Week().Thursday().At("02:00").Do(func() {})
s.Every(1).Week().Saturday().At("02:00").Do(func() {})
```

### Additional context

Perhaps scheduledWeekday becomes a slice, and then we figure out the right way to support that in calculating the duration to next run

```golang
// Job struct stores the information necessary to run a Job
type Job struct {
	...
	scheduledWeekday  *time.Weekday // Specific day of the week to start on
	...
}
```

A custom struct with the 7 weekdays as bools might be a good option.

joaolcaas:
Hi guys, how are you? Is this issue available still? I would like to crack it up

JohnRoesler:
@joaolcaas yes, that would be wonderful!

PR #164, opened 2021-04-26
Multiple days week

joaolcaas:
### What does this do?
This PR aims to add a list of weekdays instead of only one day per week in a weekly schedule. Basically, now a `Job` has a list of weekdays where the schedule can pass through and see which day is the next day in the schedule, given the job days.

Solves #127 

Yes, I have added 3 new tests and modified one (if there is any problem, we can check this out)

I've added some comments and doc changing

- [ ] Updated `README.md`


example_test.go
```diff
@@ -131,6 +131,10 @@ func ExampleJob_Weekday() {
 	// Output:
 	// Monday
 	// job not scheduled weekly on a weekday
+
+	secondWeeklyJob, _ := s.Every(1).Week().Monday().Friday().Do(task)
+	weeldays, _ := secondWeeklyJob.Weekday()
```
JohnRoesler, line 136:
```suggestion
	weekdays, _ := secondWeeklyJob.Weekday()
```

example_test.go
```diff
@@ -131,6 +131,10 @@ func ExampleJob_Weekday() {
...
+	fmt.Println(weeldays)
```
JohnRoesler, line 137:
```suggestion
	fmt.Println(weekdays)
```

job.go
```diff
@@ -14,21 +14,21 @@ import (
...
+	interval          int             // pause interval * unit between runs
+	duration          time.Duration   // time duration between runs
+	unit              schedulingUnit  // time units, ,e.g. 'minutes', 'hours'...
+	startsImmediately bool            // if the Job should run upon scheduler start
+	atTime            time.Duration   // optional time at which this Job runs when interval is day
+	startAtTime       time.Time       // optional time at which the Job starts
+	error             error           // error related to Job
+	lastRun           time.Time       // datetime of last run
+	nextRun           time.Time       // datetime of next run
+	scheduledWeekday  []*time.Weekday // Specific the days of the week to start on
```
JohnRoesler, line 26:
```suggestion
	scheduledWeekday  []*time.Weekday // Specific days of the week to start on
```

job.go
```diff
@@ -177,11 +177,18 @@ func (j *Job) ScheduledAtTime() string {
 
 // Weekday returns which day of the week the Job will run on and
 // will return an error if the Job is not scheduled weekly
-func (j *Job) Weekday() (time.Weekday, error) {
+func (j *Job) Weekday() ([]time.Weekday, error) {
```
JohnRoesler, line 180: rather than changing the signature of this method, could we add `Weekdays()` and have that return the slice.

Then, have this method return the first weekday in the slice to maintain the old behavior of someone having only a single weekday scheduled?

JohnRoesler (commented):
great stuff! some minor comments, checkout the golangci results, and the comment about function naming

job.go
joaolcaas: Done!

example_test.go
joaolcaas: the test that involves this typo will be moved to another function

example_test.go
joaolcaas: the test that involves this typo will be moved to another function

job.go
joaolcaas: Sure, that can be done!

joaolcaas: I will also take a look into golangci

job.go
```diff
@@ -14,21 +14,21 @@ import (
...
+	interval          int             // pause interval * unit between runs
+	duration          time.Duration   // time duration between runs
+	unit              schedulingUnit  // time units, ,e.g. 'minutes', 'hours'...
+	startsImmediately bool            // if the Job should run upon scheduler start
+	atTime            time.Duration   // optional time at which this Job runs when interval is day
+	startAtTime       time.Time       // optional time at which the Job starts
+	error             error           // error related to Job
+	lastRun           time.Time       // datetime of last run
+	nextRun           time.Time       // datetime of next run
+	scheduledWeekday  []*time.Weekday // Specific days of the week to start on
```
Streppel, line 26: Now that we have a slice here I don't think we still need to be working with pointers anymore, what do you think? That'll make implementation simpler and safer.

job.go
```diff
@@ -181,7 +181,23 @@ func (j *Job) Weekday() (time.Weekday, error) {
...
+	return *j.scheduledWeekday[0], nil
+}
+
+// Weekdays returns a slice of time.Weekday that the Job will run in a week and
+// will return an error fi the Job is not shceduled weekly
+func (j *Job) Weekdays() ([]time.Weekday, error) {
+	var jobWeekDays []time.Weekday
+
+	if j.scheduledWeekday == nil {
+		jobWeekDays = append(jobWeekDays, time.Sunday)
```
Streppel, line 193: Out of curiosity, why do we append Sunday and return the slice plus the err here? If the job is not scheduled for a weekly basis maybe we could get a nil slice.

job.go
```diff
@@ -181,7 +181,23 @@ func (j *Job) Weekday() (time.Weekday, error) {
 	if j.scheduledWeekday == nil {
 		return time.Sunday, ErrNotScheduledWeekday
 	}
-	return *j.scheduledWeekday, nil
+	return *j.scheduledWeekday[0], nil
+}
+
+// Weekdays returns a slice of time.Weekday that the Job will run in a week and
+// will return an error fi the Job is not shceduled weekly
```
Streppel, line 188:
```suggestion
// will return an error if the Job is not shceduled weekly
```

Streppel:
Great stuff indeed, thanks for your contribution @joaolcaas! Left some minor comments as well, let me know what you think.

job.go
joaolcaas: Solved!

job.go
joaolcaas: I may be wrong about this, but this error was never called before. To calculate the [remaining days](https://github.com/go-co-op/gocron/blob/master/scheduler.go#L233) the scheduler just use the job attribute `scheduledWeekday`. I was reluctant to change this piece but that might a good idea. Will not break the current behavior.

joaolcaas: *just for test purposes

job.go
joaolcaas: Instead of check if `scheduledWeekday` equals `nil`, now I just check the length. Done!

job.go
joaolcaas: Done!

joaolcaas:
> Great stuff indeed, thanks for your contribution @joaolcaas! Left some minor comments as well, let me know what you think.
> 
> One final food for thought: what does `s.Every(2).Weekday(thursday, friday)` (sic) means? I'd say "every 2 weeks, at Thursday and Friday", so after a friday run it'd schedule the job for next Thursday in 2 weeks (then for the next day, friday). Do you guys agree?

I think that's right, I checked the current behavior and this is almost what you said, with the difference that the schedule starts the next week when the interval is `2`. Then it back to the week jump

Streppel:
@joaolcaas indeed if we're to "fix" this it is out of scope of this PR. WDYT @JohnRoesler?

job.go
joaolcaas: Just let the `Weekday()` with this berravior. Done!

JohnRoesler:
> @joaolcaas indeed if we're to "fix" this it is out of scope of this PR. WDYT @JohnRoesler?

Agreed. We should go ahead and open an issue though so we know it needs addressing.

JohnRoesler (changes requested):
just some linting errors to address

JohnRoesler (approved):

PR #173, opened 2021-05-06
Fix calculatemonths()

### LINKED ISSUE #172: [BUG] - calculateMonths() calculating wrong duration
### Describe the bug

calculateMonths() is returning the wrong duration in the following cases:
1. If job.dayOfTheMonth is < than current date and job.atTime is > the current time.
2. If job.dayOfTheMonth is > than current date and job.atTime is < the current time.

In case 1 the nextRun date gets set to job.dayOfTheMonth+1.
In case 1 the nextRun date gets set to job.dayOfTheMonth-1.

In both cases the job will run continually due to wrong date and difference calculations.

### To Reproduce

1. Assuming current time and date as: 2021-05-06 05:00:00 
2a. s.Every(1).Month(2).At("12:00").Do(task))
2b. s.Every(1).Month(10).At("02:00").Do(task))

### Version

v1.5.0

### Expected behavior

nextRun date should not roll over nor the job should run continually on the specified cases.

joseccandido:
### What does this do?
Switches to using raw duration from a day-clamped duration for the difference between lastRun and jobDay.

Resolves #172

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -3,7 +3,6 @@ package gocron
 import (
 	"context"
 	"fmt"
-	"math"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


JohnRoesler:
@joseccandido could you add tests that highlight the failure you discovered in the issue? You can use the fake time interface to get the scheduler on exactly the time you want. Ideally, the test should fail without your commit and then pass with it.

joseccandido:
@JohnRoesler Added some tests. Please see if they suffice. As you mentioned, they pass with the commit and fail without.

JohnRoesler (approved):
thank you @joseccandido! nice PR

PR #182, opened 2021-05-31
Clear: remove also tags from scheduler

Dasio:
### What does this do?
If scheduler is started with `TagsUnique()`, tags are cleared also after Clear.

No.

No. It's bug fix.


JohnRoesler (approved):
Thank you!

PR #189, opened 2021-06-09
Optimize computing startAtTime in the future

Dasio:
### What does this do?
Optimization: instead of iterative adding duration to startAtTime, add it in one operation.

E.g if jobs startsAt year 1970 every minute. And we'll try add this job scheduler. It takes around 30s on my local machine, just to compute time in the future.

No. Existing tests works.

- [ ] Updated `README.md`


JohnRoesler (approved):
thanks for improving this logic!

Issue #210, opened 2021-08-12
[BUG] - Job deletion and addition fails when the job was started with LimitRunsTo(1).


indranil-elisity:
### Describe the bug

I am adding a job with tag 'oneTimeJob' and using option LimitRunsTo(1). The package documentation says that the job will run only once but will stay on scheduler. However, once the job has run I can neither delete the job nor I can add a new job with same tag. Please see the code below.

### To Reproduce

```
package main

import (
        "bufio"
        "fmt"
        "os"
        "time"

        "github.com/go-co-op/gocron"
)

var s1 *gocron.Scheduler

func taskWithParams(a int, b string, isMono bool) {
    fmt.Println(a, b)
    t := time.Now()
    fmt.Printf("Cron taskWithParams tag %s ran at %s\n", b, t.Local())
    if isMono == true {
        fmt.Println("This is an one time job")
    }
}

func main() {
    t := time.Now()
    fmt.Printf("Go launched at %s\n", t.Local())

    // defines a new scheduler that schedules and runs jobs
    s1 = gocron.NewScheduler(time.Local)
    s1.TagsUnique()
    s1.StartAsync()

    scanner := bufio.NewScanner(os.Stdin)

    for {
        fmt.Print("Enter add for adding job, rem for removing job, run for running a job: ")
        var option string
        var cronString string
        var tag        string
        scanner.Scan()
        option = scanner.Text()
        switch option {
        case "add":
                fmt.Print("Enter crontab: ")
                scanner.Scan()
                cronString = scanner.Text()
                fmt.Print("Enter tag: ")
                scanner.Scan()
                tag = scanner.Text()
                job, err := s1.Cron(cronString).Tag(tag).LimitRunsTo(1).Do(taskWithParams, 1, tag, true)
                if err != nil {
                    fmt.Println(err)
                    break
                }
                t = time.Now()
                fmt.Printf("Cron job with tag %s added at %s\n", job.Tags(), t.Local())
        case "rem":
                fmt.Print("Enter tag to delete: ")
                scanner.Scan()
                tag = scanner.Text()
                err := s1.RemoveByTag(tag)
                if err != nil {
                    fmt.Println(err)
                    break
                }
                fmt.Println("Remaining number of jobs ", s1.Len())
        case "run":
                fmt.Print("Enter tag to run:" )
                scanner.Scan()
                tag = scanner.Text()
                s1.RunByTag(tag)
        default:
                fmt.Print("invalid option chosen \n")
        }
    }
[...]

PR #211, opened 2021-08-16
Fix unique tag removal

JohnRoesler:
### What does this do?

The remove functions have a bug where the unique tags set at the scheduler are only removed when using RemoveByTag which isn't used by the removal that occurs when the limit runs to is reached.

Updated the removal functions so that a check occurs to see if unique tags are being used and if so, remove all of the job's tags from the unique slice.

resolves https://github.com/go-co-op/gocron/issues/210

👍 

- [ ] Updated `README.md`

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -18,7 +18,7 @@ linters:
     - deadcode
     - errcheck
     - gofmt
-    - golint
+    - revive
     - gosec
     - gosimple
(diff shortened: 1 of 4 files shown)


djenriquez:
Awesome! Was just running into this problem!

Streppel (approved):
:clap:

PR #212, opened 2021-08-21
Replace tag map with thread safe sync.map

djenriquez:
### What does this do?
This PR replaces that `uniqueTag` map with the [sync.Map](https://pkg.go.dev/sync#Map) struct which allows safe concurrent operations against the map. Before this, scheduled jobs that executed on the same schedule were prone to causing a "concurrent map" panic.

Tests have been updated to properly use the sync.map.

### Notes
This change should produce a no-op, just fixes the potential panic from jobs scheduled at the same time that require `uniqueTags`.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -538,7 +538,7 @@ func (s *Scheduler) removeJobsUniqueTags(job *Job) {
 	if job == nil {
 		return
 	}
-	if len(job.tags) > 0 {
+	if s.tagsUnique && len(job.tags) > 0 {
 		for _, tag := range job.tags {
 			s.tags.Delete(tag)
 		}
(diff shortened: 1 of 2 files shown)


scheduler.go
```diff
@@ -637,9 +639,7 @@ func (s *Scheduler) Clear() {
 		job.stop()
 	}
 	s.setJobs(make([]*Job, 0))
-	if s.tags != nil {
-		s.TagsUnique()
-	}
+	s.TagsUnique()
```
JohnRoesler, line 642: We only want to set the tags to unique if they were originally set on the scheduler. So this should probably still have a if check to say if not false then set to an new empty sync map. Am I missing something with how this map works or anything?

djenriquez: Ah, I see, the purpose of this originally was to reinstantiate the tag map to clear out the tags. Now that it is using a `sync.Map`, it would need to go through and delete the items within.

djenriquez: Fixed and verified with TDD.

JohnRoesler (approved):

PR #232, opened 2021-09-10
Get current job every bug

### LINKED ISSUE #230: [BUG] - Scheduling a job and calling get current job before every/cron panics
### Describe the bug

When scheduling a job, if I call a func that calls `getCurrentJob()` prior to the `newJob()` func being called, a slice index panic occurs.

### To Reproduce

```go
s.WaitForSchedule().Every("1m").Do(func() {})
```

output:
```
panic: runtime error: index out of range [-1]
```

### Version

v1.7.1

### Expected behavior

the panic should be handled through a slice bounds check and surfaced to the caller

### Additional context

The solution should be generic to catch / handle an error anytime a func is called prior to the instantiation of a job. 

Couple options:

1. Continue to enforce the Every/Cron methods be called first and catch the error

2. Allow methods to be called in any order, causing a new job to be created by whichever func is called first.

I like the 2nd approach better, and actually I think a bit of redesign here to think of this like a chain could be good allowing any order and the creation of a job to occur as implicitly in the first link of the chain.

denizzengin:
### What does this do?
fix panic error and use WaitForSchedule method before new job created.

Resolve #230 

- [ x] Updated `example_test.go`
- [ ] Updated `README.md`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -643,6 +643,7 @@ func ExampleScheduler_WaitForSchedule() {
 
 	// job will run 5 minutes from the scheduler starting
 	_, _ = s.Every("5m").WaitForSchedule().Do(task)
+	_, _ = s.WaitForSchedule().Every("5m").Do(task)
 
 	// job will run immediately and 5 minutes from the scheduler starting
 	_, _ = s.Every("5m").Do(task)
--- a/scheduler.go
+++ b/scheduler.go
@@ -30,9 +30,10 @@ type Scheduler struct {
 
 	tags sync.Map // for storing tags when unique tags is set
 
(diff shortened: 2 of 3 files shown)


JohnRoesler:
hey @denizzengin - this is an interesting way to get around the specific case of calling the waitforschedule method, however, I'd like to have a generic way to solve the problem as there are many methods that call the getCurrentJob func and will result in a panic, for example: week, day, month, second, etc. essentially any func called before Every() or Cron() will cause a panic.

It will take some redesigning as a lot of those methods don't return an error directly. I will give it some more thought on put some comments in the issue around this.

PR #234, opened 2021-09-12
Fix panic err job create before set behaviour.

### LINKED ISSUE #230: [BUG] - Scheduling a job and calling get current job before every/cron panics

denizzengin:
### What does this do?
fix panic error  and can be set behaviour before new job created.

Resolve #230

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -981,7 +981,7 @@ func (s *Scheduler) Sunday() *Scheduler {
 }
 
 func (s *Scheduler) getCurrentJob() *Job {
-	if len(s.jobs) == 0 {
+	if len(s.Jobs()) == 0 {
 		job := &Job{}
 		s.setJobs(append(s.Jobs(), job))
 		s.isSetBehaviourBeforeJobCreated = true
--- a/scheduler.go
+++ b/scheduler.go
@@ -30,9 +30,10 @@ type Scheduler struct {
 
 	tags sync.Map // for storing tags when unique tags is set
 
-	tagsUnique      bool // defines whether tags should be unique
-	updateJob       bool // so the scheduler knows to create a new job or update the current
-	waitForInterval bool // defaults jobs to waiting for first interval to start
+	tagsUnique                     bool // defines whether tags should be unique
+	updateJob                      bool // so the scheduler knows to create a new job or update the current
+	waitForInterval                bool // defaults jobs to waiting for first interval to start
+	isSetBehaviourBeforeJobCreated bool // represents if set some feature before job created
 }
 
 // days in a week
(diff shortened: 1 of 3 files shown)


denizzengin:
hi @JohnRoesler , 
What do you think above solution ? 
it's more generic than  my first solution.

scheduler.go
```diff
@@ -977,6 +981,11 @@ func (s *Scheduler) Sunday() *Scheduler {
 }
 
 func (s *Scheduler) getCurrentJob() *Job {
+	if len(s.jobs) == 0 {
```
JohnRoesler, line 984:
```suggestion
	if len(s.Jobs()) == 0 {
```

scheduler.go
```diff
@@ -977,6 +981,11 @@ func (s *Scheduler) Sunday() *Scheduler {
...
+		job := &Job{}
+		s.setJobs(append(s.Jobs(), job))
```
JohnRoesler, line 986:
```suggestion
		s.setJobs([]*Job{{}})
```

scheduler.go
```diff
@@ -30,9 +30,10 @@ type Scheduler struct {
 
 	tags sync.Map // for storing tags when unique tags is set
 
-	tagsUnique      bool // defines whether tags should be unique
-	updateJob       bool // so the scheduler knows to create a new job or update the current
-	waitForInterval bool // defaults jobs to waiting for first interval to start
+	tagsUnique                     bool // defines whether tags should be unique
+	updateJob                      bool // so the scheduler knows to create a new job or update the current
+	waitForInterval                bool // defaults jobs to waiting for first interval to start
+	isSetBehaviourBeforeJobCreated bool // represents if set some feature before job created
```
JohnRoesler, line 36:
```suggestion
	tagsUnique      bool // defines whether tags should be unique
	updateJob       bool // so the scheduler knows to create a new job or update the current
	waitForInterval bool // defaults jobs to waiting for first interval to start
	jobCreated      bool // so the scheduler knows a job was created prior to calling Every or Cron
```

JohnRoesler: @denizzengin if you can make this change and refactor all the references, then I'll approve!

denizzengin: @JohnRoesler  it's done but there is a test error but i run local tests successfully - (with make test)
Actually, my changes don't releated with TestScheduler_SetMaxConcurrentJobs method but gave error. i don't know what's going on here. Any suggestion ?

JohnRoesler (approved):

scheduler.go
JohnRoesler: occaisionally tests fail due to timing / slowness in the docker containers that actions uses. I've been meaning to open an issue for us to refactor the tests to use a minimum of 1 second duration because I have some that use milliseconds interval in testing and it's not reliable in the actions pipeline

JohnRoesler:
@denizzengin thanks for the contribution! ❤️

PR #238, opened 2021-09-13
fix: stopping scheduler waits for all jobs to finish

### LINKED ISSUE #237: [BUG] - Scheduler should wait all jobs to finish before returning .Stop()
### Describe the bug

Scheduler should wait all jobs to finish before returning .Stop()

### To Reproduce

```go
i := 0
s := NewScheduler(time.UTC)
s.Every(1).Second().Do(func() {
	time.Sleep(2 * time.Second)
	i = 1
})
s.StartAsync()
time.Sleep(1 * time.Second) // time to ignite job
s.Stop()
// on this line variable i should be 1 
```

### Version

v1.8.0

Streppel:
### What does this do?

Scheduler now waits for all running jobs to finish before returning. Fixes #237. 

Fixes #237 

None.

Yes.

Yes.
- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -26,28 +26,31 @@ const (
 
 type executor struct {
 	jobFunctions   chan jobFunction
-	stop           chan struct{}
+	stopCh         chan struct{}
 	limitMode      limitMode
 	maxRunningJobs *semaphore.Weighted
+
+	// runningJobsWg represents all running jobs at a current moment
+	runningJobsWg *sync.WaitGroup
 }
 
 func newExecutor() executor {
 	return executor{
-		jobFunctions: make(chan jobFunction, 1),
-		stop:         make(chan struct{}, 1),
+		jobFunctions:  make(chan jobFunction, 1),
+		stopCh:        make(chan struct{}, 1),
+		runningJobsWg: &sync.WaitGroup{},
 	}
 }
 
 func (e *executor) start() {
-	wg := sync.WaitGroup{}
 	stopCtx, cancel := context.WithCancel(context.Background())
 
 	for {
 		select {
 		case f := <-e.jobFunctions:
-			wg.Add(1)
+			e.runningJobsWg.Add(1)
 			go func() {
-				defer wg.Done()
+				defer e.runningJobsWg.Done()
 
 				if e.maxRunningJobs != nil {
 					if !e.maxRunningJobs.TryAcquire(1) {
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -698,7 +699,7 @@ func (s *Scheduler) Stop() {
 
 func (s *Scheduler) stop() {
 	s.setRunning(false)
-	s.executor.stop <- struct{}{}
+	s.executor.stop()
```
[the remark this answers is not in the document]
Streppel: oh, this is the .stop() from executor, not the scheduler itself!

JohnRoesler (approved):
very nice

scheduler_test.go
```diff
@@ -631,6 +631,19 @@ func TestScheduler_Stop(t *testing.T) {
 		s.Stop()
 		assert.False(t, s.IsRunning())
 	})
+	t.Run("waits for jobs to finish processing before returning .Stop()", func(t *testing.T) {
+		t.Parallel()
+		i := 0
+		s := NewScheduler(time.UTC)
+		s.Every(1).Second().Do(func() {
+			time.Sleep(2 * time.Second)
+			i = 1
```
JohnRoesler, line 640: this is creating a race condition. you could as atomic int, or mutex for I here, or our semaphore pattern

JohnRoesler: or no, this has to do with the stop method

executor.go
```diff
@@ -92,10 +95,14 @@ func (e *executor) start() {
...
-		case <-e.stop:
+		case <-e.stopCh:
 			cancel()
-			wg.Wait()
 			return
 		}
 	}
 }
+
+func (e executor) stop() {
```
JohnRoesler, line 105: any reason for this to be a non-pointer `executor` vs `e *executor`?

scheduler_test.go
```diff
@@ -631,6 +631,19 @@ func TestScheduler_Stop(t *testing.T) {
...
+		// i := 0
+		s := NewScheduler(time.UTC)
+		s.Every(1).Second().Do(func() {
+			time.Sleep(2 * time.Second)
+			// i = 1
+		})
+		s.StartAsync()
+		time.Sleep(1 * time.Second) // enough time for job to run
+		s.Stop()
+		assert.Equal(t, 1, 1)
```
JohnRoesler, line 645: 1 will always equal 1 lol

JohnRoesler (approved):

PR #249, opened 2021-10-31
added tests for executor and gocron

GaruGaru:
### What does this do?
Add tests for executor and gocron reflection


JohnRoesler (approved):
very nice contribution to the test base, thank you!

PR #253, opened 2021-11-07
Append tags rather than overwriting

calvinmclean:
### What does this do?
This PR changes `scheduler.Tag()` to append to, rather than overwrite, existing tags

I didn't create an issue because the README mentioned small changes can just send a PR. Here are some details I would have put in the issue:

#### Describe Bug
If multiple calls to `scheduler.Tag()` are chained together, only the last one called persists.

#### How To Reproduce
Check out [this Playground](https://play.golang.org/p/8ozf_TpCppd)

#### Expected Behavior
I expected multiple calls to `Tag()` to append to existing tags rather than overwrite.

This shouldn't be a breaking change unless someone is using multiple calls to `Tag()` to conditionally overwrite previous ones.

Yes

No

- [ ] Updated `README.md`

### Notes
While this was unexpected behavior to me, I understand it might be the intended way to use Tags, in which case this can be closed.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -788,7 +788,7 @@ func (s *Scheduler) Tag(t ...string) *Scheduler {
 		}
 	}
 
-	job.tags = t
+	job.tags = append(job.tags, t...)
 	return s
 }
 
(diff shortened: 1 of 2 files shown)


JohnRoesler (approved):
Thank you

PR #263, opened 2021-12-02
Allow removing Jobs that match multiple tags

calvinmclean:
### What does this do?
This changes `RemoveByTag` to accept multiple tags to match. All tags must be matched by the Job to be removed.

N/A

These changes should not break any current functionality because the function name is unchanged and variadic parameter allows one or more inputs.

Yes

- [ ] Updated `README.md`

### Notes
If it would be better to rename the function to `RemoveByTags`, I can do that, but I didn't want to make a breaking change.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -619,9 +619,9 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
 	s.setJobs(retainedJobs)
 }
 
-// RemoveByTag will remove a job by a given tag.
-func (s *Scheduler) RemoveByTag(tag string) error {
-	jobs, err := s.findJobsByTag(tag)
+// RemoveByTag will remove Jobs that match all given tags.
+func (s *Scheduler) RemoveByTag(tags ...string) error {
+	jobs, err := s.findJobsByTag(tags...)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


scheduler.go
```diff
@@ -619,9 +619,9 @@ func (s *Scheduler) removeByCondition(shouldRemove func(*Job) bool) {
 	s.setJobs(retainedJobs)
 }
 
-// RemoveByTag will remove a job by a given tag.
-func (s *Scheduler) RemoveByTag(tag string) error {
-	jobs, err := s.findJobsByTag(tag)
+// RemoveByTag will remove Jobs that match all given tags.
+func (s *Scheduler) RemoveByTag(tags ...string) error {
```
JohnRoesler, line 623: Will you add `RemoveByTags` as a separate function and then have it simply call `RemoveByTag`. Just provides the same grammatical syntax that the library follows for `Second` and `Seconds` e.g.

JohnRoesler:
Great contribution! Thank you so much!

scheduler.go
calvinmclean: Good idea, I'll add that 👍

JohnRoesler (approved):
👍 🎉

PR #268, opened 2021-12-11
Make FindJobsByTag exported

calvinmclean:
### What does this do?
This exports the `FindJobsByTag` function so that it can be used outside of the package. This is really useful for updating `Jobs` or displaying the `NextRun` time.

N/A

None (unless this was un-exported for a reason).

No because there is full coverage from this being used throughout the package. Let me know if I should add some specific tests.

- [ ] Updated `README.md`


JohnRoesler (approved):
👍

PR #298, opened 2022-02-17
fix weekday duration calculation error

### LINKED ISSUE #171: [BUG] - Weekday() starts schedule on next day even when current day is set
### Describe the bug

While using Weekday(), the schedule starts on the following day even if the current day is set.
eg.: If the current Day is Sunday and Sunday+Monday+Tuesday are set, the first run will be scheduled for Monday.

### To Reproduce

1. s.Every(1).Weekday(time.Sunday).Weekday(time.Monday).Weekday(time.Tuesday)
or
1. s.Every(1).Sunday().Monday().Tuesday()

### Version

1.50

### Expected behavior

If the current day is part of the weekdays set, the schedule should begin on the current day and not on the next desired weekday.

xwjdsh:
### What does this do?
fix #171

Yes.

### Notes
`remainingDaysToWeekday` is a bit confusing, I rewrite it.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -306,7 +306,7 @@ func calculateNextRunForMonth(s *Scheduler, job *Job, lastRun time.Time, dayOfMo
 }
 
 func (s *Scheduler) calculateWeekday(job *Job, lastRun time.Time) nextRun {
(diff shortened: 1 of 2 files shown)


JohnRoesler (approved):
👍

PR #307, opened 2022-03-05
add DoWithJobDetails passes a copy of the job to the jobFunc

### LINKED ISSUE #251: [BUG] - Makefile commands are out of date
### Describe the bug

> * `make test` appears to be non-deterministic in `TestScheduler_Update` and sometimes fails in `happy_path` or `update,_delete,_create`, or it simply passes. As far as I can tell, this PR did no changes in the code paths for these tests.
> * `make lint` (although `golint` has been [deprecated](https://github.com/golang/lint#readme)) outputs errors unrelated to this PR:
> ```
> $ make lint
> ~/gocron/scheduler.go:1054:1: exported method Scheduler.Cron should have comment or be unexported
> ~/scm/gocron/scheduler.go:1058:1: exported method Scheduler.CronWithSeconds should have comment or be unexported
> Found 2 lint suggestions; failing.
> make: *** [Makefile:27: lint] Error 1
> ```
> * `make vet` passes
> * `make fmt` has been run

### To Reproduce

1. A
1. B
1. C

### LINKED ISSUE #299: [FEATURE] - Add new DoWithJob() function

I have a use case where, inside the `Do()` task function, I need to know when the next scheduled job will be for the job the task is running on. Specifically (and outside the scope of gocron) because I want to set an appropriate TTL value for an MQ message being created for the job/task at hand so that it will automatically expire before the next job is scheduled if it hasn't been pulled off the queue already.

I'm certainly open to better suggestions/names, but I think something like a `DoWithJobDetails()` function very similar to `Do()` but passes a `*Job` as an argument to the supplied `jobFun` function. Note that I believe the `*Job` provided to the function will need to be sure not to be a pointer to the same internal `*Job` object that lives inside its scheduler and to instead be a copy of it. Otherwise there could be a race condition where the Job attributes could suddenly change at any time while you're trying to do something with the data. This is indeed the case in my hacky solution I attempted to do get the Job details from inside a `Do()` function.

For each cronjob I created in the scheduler, I tried to save them in a map and index them by some unique-identifying criteria. In my `jobFun` function, I then fetch the indexed/saved job for that job the function is running on. I then try to look at when the job will run next, but there are race conditions when doing so, because that data changes as the scheduler sets it up for its next run. I do hit the `"Impossible job expiration time!"` condition, although pretty rarely. Some example bits of code I have in my `jobFun`:
```
func task(someIdentifyingJobKey String) {
	savedCron := getCronJob(someIdentifyingJobKey) // this gives me the original *Job scheduled

	// HERE BE DRAGONS: There is a race condition here where the time of the
	// next future run of this job will either be ScheduledTime() or
	// NextRun(), depending up where the lifecycle of the job happens to be
	// at this point. If this job isn't pulled off the queue by an agent by
	// the time the next job comes around, we should expire it before that
	// future job takes its place to avoid a backed-up queue.
	var expires time.Duration
	scheduledTime := savedCron.ScheduledTime()
	nextRun := savedCron.NextRun()
	lastRun := savedCron.LastRun()
	if scheduledTime.Before(nextRun) {
		expires = time.Until(nextRun)
	} else {
		expires = time.Until(scheduledTime)
	}

	if expires <= 0 {
		fmt.Println("Impossible job expiration time!")
		fmt.Printf("now: %v\n", time.Now())
		fmt.Printf("scheduled time: %v\n", scheduledTime)
		fmt.Printf("next run: %v\n", nextRun)
		fmt.Printf("last run: %v\n", lastRun)
		return
	}
        ...
```

### Additional context

I think it makes sense for essentially a callback function to be passed the object it is a callback for--that is a common programming pattern. However, it is certainly subject to race conditions that I hopefully explained okay, hence why I think it'd have to be a complete new copy of the original Job struct when its passed into the function. Or, maybe some other structure other than a `*Job` could be passed in that just has some of the same details (ScheduledTime / NextRun / LastRun) of a Job too.

JohnRoesler:
### What does this do?
- adds a new method DoWithJobDetails that expects the passed jobFunc to have the it's final function parameter be a gocron.Job
  - an error is returned if the function doesn't meet this requirement
- changes the Job mutex to be inside a pointer struct inside the Job to allow copying the job struct not by pointer
- copy's the job parameters slice, since slices are by reference, to avoid a race condition with the executor calling the params
- move the schedule next run logic before running the job so the job knows it's next run during the current run

resolves #299 

closes #251 

👍 

- [ ] Updated `README.md`

#### main test code
```golang
package main

import (
	"github.com/go-co-op/gocron"
	"log"
	"time"
)

func task(job gocron.Job) {
	log.Printf("job last run: %s, this run: %s, job next run: %s", job.LastRun(), time.Now(), job.NextRun())
}

func main() {
	s := gocron.NewScheduler(time.Local)
	_, err := s.Tag("tag1").Every("1s").DoWithJobDetails(task)
	if err != nil {
		log.Fatalf("something went wrong: %v", err)
	}
	s.StartBlocking()
}

```

#### Output

```
2022/03/05 23:41:27 job last run: 0001-01-01 00:00:00 +0000 UTC,        this run: 2022-03-05 23:41:27.058901 -0600 CST m=+0.000511084, job next run: 2022-03-05 23:41:28.058794 -0600 CST
2022/03/05 23:41:28 job last run: 2022-03-05 23:41:27.058795 -0600 CST, this run: 2022-03-05 23:41:28.059964 -0600 CST m=+1.001576834, job next run: 2022-03-05 23:41:29.059843 -0600 CST
2022/03/05 23:41:29 job last run: 2022-03-05 23:41:28.059869 -0600 CST, this run: 2022-03-05 23:41:29.061094 -0600 CST m=+2.002707918, job next run: 2022-03-05 23:41:30.060961 -0600 CST
2022/03/05 23:41:30 job last run: 2022-03-05 23:41:29.061044 -0600 CST, this run: 2022-03-05 23:41:30.062407 -0600 CST m=+3.004022834, job next run: 2022-03-05 23:41:31.062168 -0600 CST
2022/03/05 23:41:31 job last run: 2022-03-05 23:41:30.062301 -0600 CST, this run: 2022-03-05 23:41:31.063563 -0600 CST m=+4.005180626, job next run: 2022-03-05 23:41:32.06342 -0600 CST

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,29 +5,8 @@ GO_PKGS   := $(shell go list -f {{.Dir}} ./...)
 fmt:
 	@go list -f {{.Dir}} ./... | xargs -I{} gofmt -w -s {}
 
-check-fmt:
-	@echo "Checking formatting..."
-	@FMT="0"; \
-	for pkg in $(GO_PKGS); do \
-		OUTPUT=`gofmt -l $$pkg/*.go`; \
-		if [ -n "$$OUTPUT" ]; then \
-			echo "$$OUTPUT"; \
-			FMT="1"; \
-		fi; \
-	done ; \
-	if [ "$$FMT" -eq "1" ]; then \
-		echo "Problem with formatting in files above."; \
-		exit 1; \
-	else \
-		echo "Success - way to run gofmt!"; \
-	fi
-
 lint:
-#	Add -set_exit_status=true when/if we want to enforce the linter rules
-	@golint -min_confidence 0.8 -set_exit_status $(GO_PKGS)
-
-vet:
-	@go vet $(GO_FLAGS) $(GO_PKGS)
+	@golangci-lint run
 
 test:
 	@go test -race -v $(GO_FLAGS) -count=1 $(GO_PKGS)
--- a/example_test.go
+++ b/example_test.go
@@ -261,13 +261,13 @@ func ExampleScheduler_Do() {
 	fmt.Printf("Job: %v, Error: %v", j, err)
 }
 
-func ExampleScheduler_DoWithDetails() {
+func ExampleScheduler_DoWithJobDetails() {
 	task := func(in string, job gocron.Job) {
 		fmt.Printf("this job's last run: %s\nthis job's next run: %s", job.LastRun(), job.NextRun())
 	}
 
 	s := gocron.NewScheduler(time.UTC)
-	j, err := s.Every(1).Second().DoWithDetails(task, "foo")
+	j, err := s.Every(1).Second().DoWithJobDetails(task, "foo")
 	s.StartAsync()
 	fmt.Printf("Job: %v, Error: %v", j, err)
 }
--- a/example_test.go
+++ b/example_test.go
@@ -261,6 +261,17 @@ func ExampleScheduler_Do() {
 	fmt.Printf("Job: %v, Error: %v", j, err)
 }
 
+func ExampleScheduler_DoWithDetails() {
+	task := func(in string, job gocron.Job) {
+		fmt.Printf("this job's last run: %s\nthis job's next run: %s", job.LastRun(), job.NextRun())
+	}
+
+	s := gocron.NewScheduler(time.UTC)
+	j, err := s.Every(1).Second().DoWithDetails(task, "foo")
+	s.StartAsync()
+	fmt.Printf("Job: %v, Error: %v", j, err)
+}
+
 func ExampleScheduler_Every() {
 	s := gocron.NewScheduler(time.UTC)
 	_, _ = s.Every(1).Second().Do(task)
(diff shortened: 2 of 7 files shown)


JohnRoesler:
This is missing a big piece right now. The next run isn't calculated until after the job is run. That's a key piece of information the caller might like to know. That will require some additional digging into.

Streppel (approved):

defc0n:
Thanks for this!

defc0n:
> let me know how it works for you!

Believe it is working for me and fixes my race condition when determining the next time the job will run, thanks. Took me a bit of time to realize the task function is passed as the first argument to `DoWithJobDetails()` but received as the last argument in that task function.

PR #311, opened 2022-03-10
fixes #308

### LINKED ISSUE #308: [BUG] - Every() function will create new job when input is integer in update mode
https://github.com/go-co-op/gocron/blob/20ce0b90de2ab4576cdaa79574ded5fb3d83d8aa/scheduler.go#L493
If it's update job or created job, why still create new job here? This will cause bug when the input of Every() function is an integer

For example, below code will give an exception because if it cannot find the job function in the scheduler jobs because of this bug.
```
package main

import (
	"fmt"
	"time"

	"github.com/go-co-op/gocron"
)

var task = func() { fmt.Println("hello") }

func main() {
	j, _ := s.Every(1).Second().Do(task)
	s.StartAsync()

	time.Sleep(10 * time.Second)
	_, _ = s.Job(j).Every(5).Second().Update()

	time.Sleep(30 * time.Minute)
	_, _ = s.Job(j).Every(1).Day().At("02:00").Update()
}
```

asad-urrahman:
### What does this do?
Fixes the [issue 308](https://github.com/go-co-op/gocron/issues/308)

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -490,7 +490,7 @@ func (s *Scheduler) Every(interval interface{}) *Scheduler {
 		if !(s.updateJob || s.jobCreated) {
 			job = s.newJob(interval)
 		} else {
-			job = s.newJob(interval)
+			job.interval = interval
 		}
 		if interval <= 0 {
 			job.error = wrapOrError(job.error, ErrInvalidInterval)


JohnRoesler (approved):
I was digging into this more and I think there are some other issues with Update. But this definitely fixes this particular issue. Thank you!

Issue #305, opened 2022-02-28
[FEATURE] - Run job at random interval


spatocode:
### Describe the solution you'd like

A job should be able to be run at random interval. I'm not sure the best way to achieve this but, we can do this by changing ```.Every()``` to a variadic function. For example:

scheduler.Every(1, 3, 5, 2).Hour().Do(func() {})
This randomly select any of the int arg.

OR

scheduler.Random(3, 1).Hour().Do(func() {})
This randomly runs between 1-3 hours, where 3 is the maximum number and 1 is the minimum.

JohnRoesler:
This should be done by adding a func Random as you propose. I would have it accept two ints for the range, min, max. Then during each scheduling of the job it would have to check if it's random and set a new random interval value from the range for the next run. I would also force it to round to the nearest whole number and make the range inclusive. So if it was 1, 5 hours, then the job would run at 1, 2, 3, 4, 5 hour intervals.

```golang
func (s *scheduler) Random(min, max int) *scheduler {
}
```

Do you agree? @spatocode

spatocode:
> Do you agree? @spatocode

@JohnRoesler Yea, that makes sense. Could you assign the issue to me so i could work on it?

AlexanderSutul:
Hey, @JohnRoesler ! I noticed that this issue is not solved yet. Maybe, you could assign to to me? I'd finish this during the next week. If it's still required to be added.

PR #339, opened 2022-06-07
add EveryRandom for random interval

JohnRoesler:

#305 

- [ ] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -3,6 +3,7 @@ package gocron
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"sort"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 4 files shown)


job.go
```diff
@@ -16,7 +17,8 @@ import (
 type Job struct {
 	mu *jobMutex
 	jobFunction
-	interval          int             // pause interval * unit between runs
+	interval          int             // interval * unit between runs
+	random                            // details for randomness
```
AlexanderSutul, line 21: Maybe better to add it as `random random // details for randomness` because otherwise we can use fields from `random struct` directly from `job struct`.

AlexanderSutul:
One small comment from my side. Otherwise 👍🏻

AlexanderSutul (changes requested):
One small comment from my side. Otherwise 👍🏻

PR #354, opened 2022-07-18
custom time provider for scheduler closes #353

### LINKED ISSUE #353: [FEATURE] - allow users to provide a custom timeWrapper for the scheduler

I need to test an app using gocron, and I'd like to run it using a simulated/accelerated time provider replacing the system time

I've seen that there is a timeWrapper interface already implemented in the code, but it's private. Making it public, and creating an alternate constructor NewSchedulerCustomTime(loc *time.Location, tw TimeWrapper) for the scheduler  would solve the issue -->

A solution I considered (not involving changes to gocron) is to write a function to accelerate the cronstrings directly before adding jobs to gocron. But that looks extremely tricky to implement without restricting the allowed expressions. 
I considered 2 small variant of the solution I proposed: 
- exposing the "time" field of the Scheduler stuct as public
- modifying NewScheduler directly instead of creating an alternate constructor

### Additional context

I'm providing a fork implementing my proposed solution

robertofratello:
### What does this do?
Makes timeWrapper interface public (calling it TimeWrapper) and implements alternate Scheduler constructor NewSchedulerCustomTime receiving a TimeWrapper as an argument.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -26,7 +26,7 @@ type Scheduler struct {
 	runningMutex  sync.RWMutex
 	running       bool // represents if the scheduler is running at the moment or not
 
-	time     timeWrapper // wrapper around time.Time
+	time     TimeWrapper // wrapper around time.Time
 	executor *executor   // executes jobs passed via chan
 
 	tags sync.Map // for storing tags when unique tags is set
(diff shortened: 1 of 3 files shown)


JohnRoesler:
@robertofratello how about if instead of creating a new constructor, we add a method to the existing scheduler struct to accept time.

In the tests we are currently accessing the private `time` variable on the struct and set it with the time wrapper, but I am not opposed to adding a setter for this.

Perhaps

```golang
func (s *Scheduler) CustomTime(customTimeWrapper TimeWrapper) {
	s.time = customTimeWrapper
}
```

JohnRoesler:
@robertofratello would you like to update this PR with that? I'm happy to do it, but I also would like you to get the contribution if you would like to.

I'd want 2 things:
- the new func
- an example in `example_test.go` showing how it works

PR #388, opened 2022-10-20
if job chains are used improperly we can end up with jobs with nil functions

JohnRoesler:
### What does this do?
removes the job if one is created without a function and tried to run

relates #376 

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -107,7 +107,6 @@ func (e *executor) start() {
 							return nil, nil
 						default:
 						}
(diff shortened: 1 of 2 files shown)


executor.go
```diff
@@ -106,6 +107,7 @@ func (e *executor) start() {
 							return nil, nil
 						default:
 						}
+						log.Printf("executor jobFunction: %+v\n", f)
```
JohnRoesler, line 110:
```suggestion
```

PR #392, opened 2022-10-26
fix logic to wait for executor stop

### LINKED ISSUE #391: [BUG] - Stop func might not wait for job to finish
### Describe the bug

(*Scheduler).Stop func should wait for all jobs to finish due to its [doc](https://pkg.go.dev/github.com/go-co-op/gocron#Scheduler.Stop), but it possibly returns before all jobs finished.

### To Reproduce

(This is a concurrent issue and a bit hard to reproduce)

1. Add `time.Sleep(time.Second)` after the lines below (to raise the concurrent issue on purpose)
https://github.com/go-co-op/gocron/blob/e3151123a8a5705f4dc267e7ca0e92a6584b9ba4/executor.go#L47-L48

1. Write some code like this:

```go
func main() {
	s.Every(1).Second().Do(func() {
		fmt.Println("Job started")
		time.Sleep(5 * time.Second)
		fmt.Println("Job finished")
	})
	go func() {
		time.Sleep(3 * time.Second)
		fmt.Println("Trying to stop")
		s.Stop()
		fmt.Println("Stopped")
	}()
	s.StartAsync()

	time.Sleep(10 * time.Second)
}
```

The output on my machine is:

```
Job started
Job started
Trying to stop
Stopped
Job started
Job finished
Job finished
Job finished
```

As we can see, `Stop()` func returns before the jobs finished. Expected behavior is that `Stopped` log is shown at the end.

### Version

latest main: [e315112](https://github.com/go-co-op/gocron/commit/e3151123a8a5705f4dc267e7ca0e92a6584b9ba4)

### Expected behavior

`Stop()` func always waits for the jobs to finish.

### Additional context

N/A

hiroebe:
### What does this do?
Add `stoppedCh` aside from `stopCh`, which is used to wait for the executor to stop

Resolves #391.

No

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -27,6 +27,7 @@ const (
 type executor struct {
 	jobFunctions   chan jobFunction
 	stopCh         chan struct{}
+	stoppedCh      chan struct{}
 	limitMode      limitMode
 	maxRunningJobs *semaphore.Weighted
 }
@@ -34,7 +35,8 @@ type executor struct {
 func newExecutor() executor {
 	return executor{
 		jobFunctions: make(chan jobFunction, 1),
-		stopCh:       make(chan struct{}, 1),
+		stopCh:       make(chan struct{}),
+		stoppedCh:    make(chan struct{}),
 	}
 }
 
@@ -113,13 +115,13 @@ func (e *executor) start() {
 		case <-e.stopCh:
 			cancel()
 			runningJobsWg.Wait()
-			e.stopCh <- struct{}{}
+			close(e.stoppedCh)
 			return
 		}
 	}
 }
 
 func (e *executor) stop() {
-	e.stopCh <- struct{}{}
-	<-e.stopCh
+	close(e.stopCh)
+	<-e.stoppedCh
 }
--- a/executor_test.go
+++ b/executor_test.go
@@ -26,12 +26,12 @@ func Test_ExecutorExecute(t *testing.T) {
 		runState:   &runState,
 	}
 
-	e.stop()
 	wg.Wait()
+	e.stop()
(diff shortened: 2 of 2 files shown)


executor.go
```diff
@@ -27,14 +27,16 @@ const (
...
+	stoppedCh      chan struct{}
 	limitMode      limitMode
 	maxRunningJobs *semaphore.Weighted
 }
 
 func newExecutor() executor {
 	return executor{
 		jobFunctions: make(chan jobFunction, 1),
-		stopCh:       make(chan struct{}, 1),
+		stopCh:       make(chan struct{}),
```
hiroebe, line 38: We don't need capacity for this channel. Closing the channel is enough to tell the stop.

executor.go
```diff
@@ -113,13 +115,13 @@ func (e *executor) start() {
...
 			return
 		}
 	}
 }
 
 func (e *executor) stop() {
-	e.stopCh <- struct{}{}
-	<-e.stopCh
+	close(e.stopCh)
+	<-e.stoppedCh
```
hiroebe, line 126: We need two channels for this use case. In the previous implementation, `struct{}{}` sent at L123 might be received at L124, which means `start()` loop never stops.

hiroebe:
@JohnRoesler 
I've added a commit to fix the test. [474913e](https://github.com/go-co-op/gocron/pull/392/commits/474913eed5536616a12f711bbee28c7f48d84a80)
`stop()` must be called after the `jobFunction` has started. Otherwise, `jobFunction` is never called and the test never finishes.

Issue #322, opened 2022-04-08
[BUG] - Multiple times not working


nandra:
### Describe the bug

With this test code:

```
package main

import (
	"fmt"
	"time"

	"github.com/go-co-op/gocron"
)

func main() {
	s := gocron.NewScheduler(time.UTC)
	dayOn, _ := s.Every(1).Day().At("20:31;20:34;20:36").Do(func() {
		fmt.Println("on", time.Now())
	})

	dayOff, _ := s.Every(1).Day().At("20:33;20:35;20:37").Do(func() {
		fmt.Println("off", time.Now())
	})
	s.StartAsync()

	for {
		time.Sleep(10 * time.Second)
		fmt.Println("on:", dayOn.ScheduledAtTimes(), dayOn.NextRun())
		fmt.Println("off:", dayOff.ScheduledAtTimes(), dayOff.NextRun())
	}
}
```

I get call only on first time occurrence not others (see log):
```
go run main.go
on: [20:31 20:34 20:36] 2022-04-08 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on 2022-04-08 22:31:00.000364307 +0200 CEST m=+14.149554983
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
off: [20:33 20:35 20:37] 2022-04-08 20:33:00 +0000 UTC
on: [20:31 20:34 20:36] 2022-04-09 20:31:00 +0000 UTC
[...]

PR #393, opened 2022-11-02
Remove the code that double check last run time and current time

xwjdsh:
### What does this do?
https://github.com/go-co-op/gocron/issues/322#issuecomment-1301104591

#322 

No breaking changes.


Streppel (approved):
@xwjdsh thank you so much for taking the time to investigate #322. This is the beauty of open source. Everything looks good from my side after some manual testing. @JohnRoesler please take a look at this too!

PR #413, opened 2023-01-30
Fix flaky SetEventListeners test

zhsj:
Stop should be called after the job has been scheduled.

The failure can be reproduced by

```
go test -p 4 -count=50 -v -run TestJob_SetEventListeners
```


JohnRoesler:
thank you!

PR #417, opened 2023-02-03
Add telemetry

### LINKED ISSUE #317: [FEATURE] - Add Prometheus Metrics For Job
We can add prometheus metrics for job execution.Then outside monitor system can use these metrics to track jobs.
Maybe I can send a pr.

arjunmahishi:
### What does this do?
This PR adds some useful metrics to observe the rate and performance of scheduled jobs

resolves #317 

N/A

- [x] Updated `README.md`

### Notes

**Metrics being emitted**

```go
jobLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
	Name:    "gocron_job_latency_milliseconds",
	Help:    "time taken to execute a job in milliseconds",
	Buckets: LatencyBuckets,
}, []string{"job_name", "scheduler_name"})
```

This is an opt-in feature. The metrics are only emitted if the user explicitly enables it with `gocron.EnableMetrics()`.

**POC - screenshot**

image


telemetry.go
```diff
@@ -0,0 +1,31 @@
...
+	// 	Help: "The total number of job failures.",
+	// })
+
+	metricsEnabled = false
+)
+
+func init() {
+	prometheus.MustRegister(jobLatency)
+	// prometheus.MustRegister(jobErrors)
+}
```
JohnRoesler, line 21: this should check if metrics are enabled before registering the metric, and if we can't do that within the init, function, then move this inside the EnableMetrics func

telemetry.go
```diff
@@ -0,0 +1,31 @@
...
+	jobLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "gocron_job_latency_seconds",
+		Help: "The latency of jobs in seconds.",
+	}, []string{"job_name"})
+	// jobErrors = prometheus.NewCounter(prometheus.CounterOpts{
+	// 	Name: "gocron_jobs_failures_total",
+	// 	Help: "The total number of job failures.",
+	// })
+
+	metricsEnabled = false
```
JohnRoesler, line 15: I think metrics should be tied/scoped to the scheduler. If someone has multiple schedulers running, might want to have metrics enabled on some but not others.

telemetry.go
```diff
@@ -0,0 +1,31 @@
+package gocron
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	jobLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "gocron_job_latency_seconds",
+		Help: "The latency of jobs in seconds.",
+	}, []string{"job_name"})
```
JohnRoesler, line 9: might want to optionally include a job_name_prefix or scheduler_name_prefix option that could be set when enabling metrics on the scheduler to prefix all metric labels on the scheduler with something.

JohnRoesler:
@arjunmahishi the screen shot you posted looks like it recorded all 8 runs for all the vectors in the histogram? maybe I'm missing something?

arjunmahishi:
> @arjunmahishi the screen shot you posted looks like it recorded all 8 runs for all the vectors in the histogram? maybe I'm missing something?

Yes. This metric emits the histogram buckets. Each bucket has a `le` label which represents the  **count** of jobs whose latency is "less or equal to" the value of the `le` label.

Example:
```
gocron_job_latency_seconds_buckets{job_name="abc", le="0.05"} 8
```

This means, there are 8 job executions that took less than or equal to 0.05 seconds to complete.

---

Since this is getting measured in seconds, all job executions are coming under all buckets. Because they are all finishing faster than the lowest bucket

For more: https://prometheus.io/docs/concepts/metric_types/#histogram

telemetry.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	jobLatency    *prometheus.HistogramVec
+	telemetryOnce sync.Once
+
+	// Latency buckets for job latency metrics. This can be overridden on the
+	// application side
+	LatencyBuckets = prometheus.ExponentialBuckets(1, 2.5, 15)
```
JohnRoesler, line 15: I think we should this metric perhaps `Duration` or `JobDuration`

telemetry.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func initTelemetry() {
+	telemetryOnce.Do(func() {
+		jobLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "gocron_job_latency_milliseconds",
+			Help:    "time taken to execute a job in milliseconds",
+			Buckets: LatencyBuckets,
+		}, []string{"job_name", "scheduler_name"})
```
JohnRoesler, line 24: I think the documentation should note how your metrics will show up if you use anonymous functions:

```go
package main

import (
	"fmt"
	"reflect"
	"runtime"
)

func myFunc() {}

func main() {
	// this is how getFunctionName() in our lib determines the name
	fmt.Println(runtime.FuncForPC(reflect.ValueOf(func() {}).Pointer()).Name())
	fmt.Println(runtime.FuncForPC(reflect.ValueOf(func() {}).Pointer()).Name())
	fmt.Println(runtime.FuncForPC(reflect.ValueOf(myFunc).Pointer()).Name())
	fmt.Println(runtime.FuncForPC(reflect.ValueOf(func() {}).Pointer()).Name())
}
```
outputs
```
main.main.func1
main.main.func2
main.myFunc
main.main.func3
```

scheduler.go
```diff
@@ -1335,3 +1341,19 @@ func (s *Scheduler) StopBlockingChan() {
...
+
+// EnableTelemetry enables the collection of metrics for the scheduler. Takes
+// the name of the scheduler to set it as a metric label
+func (s *Scheduler) EnableTelemetry(schedulerName string) {
+	initTelemetry()
+	s.isTelemetryEnabled = true
+	s.telemetryShdName = schedulerName
+}
+
+func (s *Scheduler) observeJobLatency(jobName string, latency float64) {
```
JohnRoesler, line 1353: rather than have this method on the scheduler, how about if we make a separate `telemetry` struct that both the scheduler and executor can take pointers to. Then we don't need the `isTelemetryEnabled` bool because the struct will either exist or be nil.

Right now, it feels odd to me that the executor is taking in a scheduler and then calling a metrics method on the scheduler.

telemetry.go
```diff
@@ -0,0 +1,32 @@
...
+
+		prometheus.MustRegister(jobLatency)
+	})
+}
+
+func observeJobLatency(latency float64, jobName, shdName string) {
```
JohnRoesler, line 30: I could see also using job tags in addition to the job name for labels on the metric

telemetry.go
```diff
@@ -0,0 +1,32 @@
+package gocron
```
JohnRoesler, line 1: it'd be cool to think about what other metics users would benefit from.

some i can think of:
- count of runs by job
- for the singleton / waitmode I could see having a metric for how many jobs are waiting / queued - could do this by incrementing a metric when one is queued and decrementing when it's run

bpurvisCB:
Why was this closed? @arjunmahishi

JohnRoesler:
Here is the direction we plan to go in: https://github.com/go-co-op/gocron/issues/317#issuecomment-1587653684

Issue #414, opened 2023-01-30
[BUG] - Flaky TestAt


zhsj:
### To Reproduce

```
go test -v -count=1000 -run '^TestAt$'
```

```
=== RUN   TestAt
=== RUN   TestAt/job_scheduled_for_future_hasn't_run_yet
panic: implement me

goroutine 126 [running]:
github.com/go-co-op/gocron.fakeTime.Sleep({0xc00007f110?}, 0x0?)
        /home/zsj/Workspaces/debian/package/pkg-go/golang-github-go-co-op-gocron/scheduler_test.go:30 +0x27
github.com/go-co-op/gocron.(*Scheduler).runContinuous.func1()
        /home/zsj/Workspaces/debian/package/pkg-go/golang-github-go-co-op-gocron/scheduler.go:597 +0x9b
created by time.goFunc
        /usr/lib/go-1.20/src/time/sleep.go:176 +0x32
exit status 2
FAIL    github.com/go-co-op/gocron      60.008s
```

### Version

main 01aebbbcfd7225038528a5e21ca9f0548db98348

### Expected behavior

No error.

kennykarnama:
It should be solved in: https://github.com/go-co-op/gocron/pull/419

PR #418, opened 2023-02-13
Attempt to fix flaky testAt

kennykarnama:
### What does this do?

This PR attempts to fix flaky test.

Suspect there is a resource leak (scheduler) on each test iteration in https://github.com/go-co-op/gocron/blob/main/scheduler_test.go#L231 so it will run continuous and eventually will trigger this line: https://github.com/go-co-op/gocron/blob/main/scheduler.go#L597

I tried to run using [goleak](https://github.com/uber-go/goleak) test and it is proven there is a leak.

So, calling stop on each assertion in that test case will free the resource.

```
=== RUN   TestAt/job_scheduled_for_future_hasn't_run_yet
2023/02/13 14:25:22 1970-01-01 12:01:00 +0000 UTC
2023/02/13 14:25:22 1970-01-01 12:01:00 +0000 UTC
    scheduler_test.go:262: found unexpected goroutines:
        [Goroutine 8 in state select, with github.com/go-co-op/gocron.(*executor).start on top of the stack:
        goroutine 8 [select]:
        github.com/go-co-op/gocron.(*executor).start(0xc000077860)
                C:/Users/kenny.karnama/go/src/github.com/go-co-op/gocron/executor.go:48 +0xe5
        created by github.com/go-co-op/gocron.(*Scheduler).start
                C:/Users/kenny.karnama/go/src/github.com/go-co-op/gocron/scheduler.go:89 +0x6a
        ]
```

https://github.com/go-co-op/gocron/issues/414

Call `s.Stop` on each assertion

- Added verifyLeak
- Run `go test -v -count=1000 -run '^TestAt$'` and there is no panic

-

### Notes

-

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,11 +5,13 @@ go 1.19
 require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/stretchr/testify v1.8.1
+	go.uber.org/goleak v1.2.0
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
 )
 
 require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/kr/text v0.2.0 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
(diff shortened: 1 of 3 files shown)


kennykarnama:
need further investigation since leak change detects other leaks in another test.

Will open once the issue has been resolved.

PR #432, opened 2023-03-08
job: Pad `ScheduledAtTime(s)` with leading zeroes

gechr:
I found it a bit odd that times were printed as `8:3` instead of `08:03` when using these methods, so pad them with leading zeroes for clarity.


JohnRoesler (approved):
thank you!

PR #435, opened 2023-03-15
Fix a bug with scheduling next runs on daylight savings change days

andyy-gci:
### What does this do?
This fixes a bug in how gocron schedules scheduled next day, week/weekday, month tasks. Previously, the next run was calculated by rounding the current date to midnight, adding the scheduled time ("at time" represented as a duration), and then adding the necessary day, week, or month interval. This methodology fails on daylight savings transition days. 

For example, suppose gocron was run on March 12, 2023 (first day of daylight savings time) and needs to set up a task that runs at 8pm every night. The initial calculation of the next run date time will calculate the next run by rounding the current time to midnight (March 12, 2023 at 00:00) and adding the scheduled time ("at time" of 20:00 represented as 20 hours in duration). Normally, this yields 8pm but since daylight savings time started earlier at 3am, the actual result is 9pm. Thus, the scheduled tasks will run 1 hour late until either a power cycle of gocron at 12:00am or after the next run on March 13, 2023.

This PR fixes the issue by removing the addition operation and instead, populate a new `time.Time` object with the last run's date and the deconstructed hours, minutes, seconds from the "at time" duration.

N/A

I don't think this breaks current functionality.

Yes

No

- [ ] Updated `README.md`

### Notes
It's possible I missed a few locations but I think I covered most of it. Please let me know if I missed any scenarios.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -239,8 +239,6 @@ func (s *Scheduler) durationToNextRun(lastRun time.Time, job *Job) nextRun {
 }
 
 func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) nextRun {
-	lastRunRoundedMidnight := s.roundToMidnight(lastRun)
-
 	// Special case: the last day of the month
 	if len(job.daysOfTheMonth) == 1 && job.daysOfTheMonth[0] == -1 {
 		return calculateNextRunForLastDayOfMonth(s, job, lastRun)
@@ -264,7 +262,7 @@ func (s *Scheduler) calculateMonths(job *Job, lastRun time.Time) nextRun {
 
 		return nextRunResult
 	}
-	next := lastRunRoundedMidnight.Add(job.getFirstAtTime()).AddDate(0, job.getInterval(), 0)
+	next := s.roundToMidnightAndAddDSTAware(lastRun, job.getFirstAtTime()).AddDate(0, job.getInterval(), 0)
 	return nextRun{duration: until(lastRun, next), dateTime: next}
 }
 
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -386,14 +387,6 @@ func in(scheduleWeekdays []time.Weekday, weekday time.Weekday) bool {
 }
 
 func (s *Scheduler) calculateDuration(job *Job) time.Duration {
-	if job.neverRan() && shouldRunAtSpecificTime(job) { // ugly. in order to avoid this we could prohibit setting .At() and allowing only .StartAt() when dealing with Duration types
```
andyy-gci, line 389: I think this [check](https://github.com/go-co-op/gocron/blob/fb8a25590a070515d6069a3dbb31a66e40c3ebdd/scheduler.go#L853) makes this dead code since `shouldRunAtSpecificTime` will never be true (`job.getAtTime(jobLastRun) != 0` and under hour unit should be already flagged by an error)

scheduler_test.go
```diff
@@ -395,6 +395,71 @@ func TestWeekdayAt(t *testing.T) {
...
+	loc, err := time.LoadLocation("US/Pacific")
+	assert.NoError(t, err)
+
+	fromRFC3339 := func(ts string) time.Time {
+		dt, err := time.Parse(time.RFC3339, ts)
+		assert.NoError(t, err)
+		return dt.In(loc)
+	}
+
+	before_to_edt := fakeTime{onNow: func(l *time.Location) time.Time {
```
JohnRoesler, line 408: could you update the vars to follow the Go camelCase preference instead of snake_case? Thanks!
```suggestion
	beforeToEDT := fakeTime{onNow: func(l *time.Location) time.Time {
```

andyy-gci: Sorry about that, should be fixed now

JohnRoesler (approved):
great fix!

PR #443, opened 2023-04-11
add func for getting all tags

jzbyers:
### What does this do?

This PR adds a function for getting all tags.

#### Use Case

Imagine you have a database table of thousands of SQL queries you need to run according to their configured crontab schedules, and you have a service that fetches these configs and uses gocron to schedule each query to execute. This service also polls the database on a schedule to sync the configured query configs with corresponding jobs in the gocron scheduler. It would be nice to have some sort of `GetAllTags()` function so that if you needed to delete jobs that no longer need to run, you can compare the list of all tags to the list of all newly fetched query configs.

```golang
scheduler := gocron.NewScheduler(time.UTC)
scheduler.TagsUnique()

// map of active queries to schedule
queryConfigs := map[string]string{
  "query1": "select * from user",
  "query2": "select * from company",
}

// remove all jobs that no longer need to run
for _, tag := range scheduler.GetAllTags() {
  if _, ok := queryConfigs[tag]; !ok {
    scheduler.RemoveByTag(tag)
  }
}
```

N/A

None

Yes

- [x] Updated `README.md` (marking this as done but didn't change anything)

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -351,6 +351,15 @@ func ExampleScheduler_Friday() {
 	// Friday
 }
 
+func ExampleScheduler_GetAllTags() {
+	s := gocron.NewScheduler(time.UTC)
+	_, _ = s.Every(1).Second().Tag("tag1").Do(task)
+	_, _ = s.Every(1).Second().Tag("tag2").Do(task)
+	fmt.Println(s.GetAllTags())
+	// Output:
+	// [tag1 tag2]
+}
+
 func ExampleScheduler_Hour() {
 	s := gocron.NewScheduler(time.UTC)
 
--- a/scheduler.go
+++ b/scheduler.go
@@ -966,6 +966,22 @@ func (s *Scheduler) Tag(t ...string) *Scheduler {
 	return s
 }
 
+// GetAllTags returns all tags.
+func (s *Scheduler) GetAllTags() []string {
+	var tags []string
+	if s.tagsUnique {
+		s.tags.Range(func(key, value interface{}) bool {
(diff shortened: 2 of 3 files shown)


scheduler.go
```diff
@@ -966,6 +966,22 @@ func (s *Scheduler) Tag(t ...string) *Scheduler {
...
 
+// GetAllTags returns all tags.
+func (s *Scheduler) GetAllTags() []string {
+	var tags []string
+	if s.tagsUnique {
+		s.tags.Range(func(key, value interface{}) bool {
+			tags = append(tags, key.(string))
+			return true
+		})
+	} else {
```
JohnRoesler, line 977: with tags unique, the tags are still set on the jobs, so I think it would be cleaner to simply range over all the jobs and join the tags.

scheduler_test.go
```diff
@@ -2,6 +2,8 @@ package gocron
 
 import (
 	"fmt"
+	"log"
```
JohnRoesler, line 5: looks like an unused import?

scheduler.go
jzbyers: 13c4482c3aa0cc460fda878c9b0aee943dc29719

example_test.go
```diff
@@ -351,6 +351,15 @@ func ExampleScheduler_Friday() {
 	// Friday
 }
 
+func ExampleScheduler_GetAllTags() {
```
JohnRoesler, line 354: thanks for adding this!

JohnRoesler (approved):

PR #460, opened 2023-04-22
fix(job): provide access to job context

cristiprg:
### What does this do?

This PR provides access to a `Job`'s `ctx` field. This comes very handy when we've got a long running job that we want to kill when the `Job` is stopped. By using the context `Done()` channel, we can exit immediately, just as the `Job` has been stopped.

This is in line with the standard practice in Go, for example https://pkg.go.dev/net/http#Request.Context.

Without this change, the user of the library has to implement quite significant code to work around this issue. Mainly maintaining a list of `context`es and/or `cancel` functions associated to each `Job`.  However, this seems redundant since all this functionality is already implemented in the library itself.

Changes:
* [main change] added a getter for the `ctx`: https://github.com/go-co-op/gocron/pull/460/files#diff-d25200def3804ad5276ed9f6b2e57f0681afe30e7b1929ccd2dacda060b920c2R297
* added test asserting the getter returns the context
* added test asserting that the cancel function is called when a job is stopped

This PR does not modify or break any new functionality, it just exposes the `ctx` of a `Job`.

Yes

- [x] Updated `README.md`
- [x] Update documentation of `Remove` method to reflect that there is a new easy way to find out whether the job has been removed/canceled


JohnRoesler (approved):
Very nice! Thank you for the contribution

Issue #181, opened 2021-05-27
[FEATURE] - Run a scheduled a job once in a distributed system


marcsantiago:
### Is your feature request related to a problem? Please describe
Currently If I use this package on any cluster of AWS ec2s, all my machines running production code would also run code in the scheduler. For instance if I need a background go routine to scan data every minute and I have 5 machines deployed. I don't need or want all my machines to scan the same data. It's also not cost effective to spin a single isolated ec2 to run a single job.

### Describe the solution you'd like
I would like if gocron had to ability to add connections to persistent storage via some sort of interface to allow users to plugin any persistent storage (redis, memcache, sql, etc) such that it kept track of the jobs running. If a job is running by 1 machine and you've set the settings to allow only 1 job globally to run, then all other machines do not run the same job. In essence it would be safe to deploy code on a fleet of web servers and have only 1 web server at any given time run the same job.

some considerations: 

- When a cluster of machines are being deployed, restarted, or added to a load balancer, there is a chance for a race condition to occur given the latency it may take the fastest instance to write to persistent storage. Example If machine A writes to storage saying "hey I'm responsible for the job" and machine B checks storage a few minutes later and doesn't find anything in persistent storage it will also say "hey I'm responsible for the job" .  This is because machines get updates in batches. In order to prevent this behavior there needs to be some sort of jitter or perhaps a queue is used.  For instance of all the servers start up and schedule a job. The name space for that job should be the same on each machine. If i'm using  redis,  I would upsert a job by some shared id. When the scheduler ticks on the application it pops from the queue, if there is an item then it's granted the ability to run the job else the job does not run. When the job completes on the machine with the successful pop, it adds the job id back to the queue for the next tick.... something like that

### Describe alternatives you've considered

- Spinning up isolated ec2s that do nothing other then run a single job, at the cost or operational cognitive load and monetary cost
- Use other distributed systems like machinery, however projects like machinery aren't necessarily meant for cron like tasks.
- Compile a Go binary and use the native linux cron system, but again this means moving away from using a very clean package and system that gocron presents

JohnRoesler:
Definitely something we’re interested in supporting

JohnRoesler:
I like the way that this platform (https://benthos.dev/) abstracts different data sources and I think it would be ideal to then add a reasonable interface to gocron that we could then support multiple specific implementations like redis, ec2, etc.

ianaz:
[...]

Issue #180, opened 2021-05-26
[FEATURE] - Run a scheduled a job once in a distributed system


marcsantiago:
It would incredibly useful to be able to cron jobs into existing server clusters but only have 1 machine in that cluster run the job. For instance if i had a fleet of AWS ec2s and I wanted my web servers to also run background jobs, I would want only 1 of those servers to run a background job, else all my machines with the same production code would execute the same tasks * N number of boxes. 

In order for this to work on a distributed system the code would have to expose an optional interface, which would allow one to use their persistent storage of choice to check whether or not the job is already running on a different machine with the same job id and some kind of check to ensure there are no race conditions on initial server deploy when all the jobs are trying to be schedule jobs at server start.

PR #463, opened 2023-04-29
add locker for distributed scheduling

JohnRoesler:
### What does this do?
adds the `WithDistributedLocker()` method to the scheduler that takes in a Locker allowing for locking jobs when multiple schedulers are running.

The initial implementation is with redis. However, additional implementations may be added here via PR or you may simply implement your own Locker. 

#181 
#180 

yes

testcontainers will soon be officially supporting redis (it's not released, but is available on main) -> https://golang.testcontainers.org/modules/redis/ 

Super awesome project!

- [x] Updated `README.md`

### Notes

This is called out in the function docs for the new method, but I will post it here also:

```
// WithDistributedLocker prevents the same job from being run more than once
// when multiple schedulers are trying to schedule the same job.
//
// One strategy to reduce splay in the job execution times when using
// intervals (e.g. 1s, 1m, 1h), on each scheduler instance, is to use
// StartAt with time.Now().Round(interval) to start the job at the
// next interval boundary.
//
// Another strategy is to use the Cron or CronWithSeconds methods as they
// use the same behavior described above using StartAt.
//
// NOTE - the Locker will NOT lock jobs using any of the limiting functions:
// SingletonMode, SingletonModeAll or SetMaxConcurrentJobs
```


locker_test.go
```diff
@@ -0,0 +1,68 @@
...
+			t.Fatalf("failed to terminate container: %s", err)
+		}
+	})
+
+	uri, err := redisContainer.ConnectionString(ctx)
+	require.NoError(t, err)
+
+	resultChan := make(chan int, 10)
+	f := func(schedulerInstance int) {
+		resultChan <- schedulerInstance
```
JohnRoesler, line 30:
```suggestion
		resultChan <- schedulerInstance
		time.Sleep(100 * time.Millisecond)
```

JohnRoesler: It runs fast enough on my Mac that the jobs all attempt to get locks at the same time. I’m guessing that within actions, the jobs are lagging slightly apart and so adding the sleep here causes the lock to stay more

executor.go
```diff
@@ -170,6 +172,15 @@ func (e *executor) run() {
 
 				switch f.runConfig.mode {
 				case defaultMode:
+					if e.distributedLocker != nil {
```
JohnRoesler, line 175: This should be moved into the scheduler so that the lock and unlock and last for the interval. If the job is set to run once a minute and the schedulers start up at slightly different times, this implementation wouldn’t block the others unless the job took a long time to run. 

One way around it would be to force jobs using the distributed locker to use StartAt to make interval jobs deterministic.

JohnRoesler: can't move it into the scheduler though because then we'd have to take into account all the different modes there. also, locking in the job continuous run function would likely result in a single scheduler running the same job over and over instead of other schedulers having a chance.

JohnRoesler:
One issue I see is that if the jobs have a very fast execution time, the current implementation only locks during the jobs execution and it's possible if the timing is slightly off on the different scheduler instances that multiple of the same job could run on multiple instances. If the jobs take some time to run then this wouldn't be a problem.

PR #471, opened 2023-05-06
Fix the distributed locking didn't take effect if the max concurrent jobs was set

git-hulk:
### What does this do?

As the title described, the distributed locking limit didn't take effect if the max concurrent jobs were set at the same time.

None

Yes

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -105,7 +105,7 @@ func (e *executor) limitModeRunner() {
 			return
 		case jf := <-e.limitModeQueue:
 			if !e.stopped.Load() {
-				runJob(jf)
+				e.runJob(jf)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


JohnRoesler:
I intentionally excluded the locker from working with the max concurrent or limit mode jobs as I’m concerned about how they will work together. 

I haven’t had a chance to try and add some more extensive testing.

Things I want to know:
- what happens when the job runs longer than interval with locker. Set to run every 1s but takes 2s to run occasionally. 
- what’s the use case for running with max concurrent as well as a locker. Max concurrent only limits how many can be running on a single scheduler and not across instances.

git-hulk:
Hi, @JohnRoesler Thanks for your reply.

> what happens when the job runs longer than interval with locker. Set to run every 1s but takes 2s to run occasionally.

Yes, this is a problem, but this situation is good for me. Maybe we can make it an option to allow to create a new job even if the last job is still running(for example, we can append the time period to the lock key).

> what’s the use case for running with max concurrent as well as a locker. Max concurrent only limits how many can be running on a single scheduler and not across instances.

From my side, we want to run the `gocron` in multi instances to improve the service availability but don't expect the same cron run to be executed more than once.

JohnRoesler:
> From my side, we want to run the gocron in multi instances to improve the service availability but don't expect the same cron run to be executed more than once.

Running multiple instances makes sense and making sure the same job doesn’t run on multiple instances at the same time.

Why do you need to limit how many different jobs can run at the same time on a single instance? As that’s what max concurrent jobs limit does.

git-hulk:
> Why do you need to limit how many different jobs can run at the same time on a single instance? As that’s what max concurrent jobs limit does.

The main purpose is to prevent too many concurrent jobs in one scheduler and cause OOM, so we will set a relatively large value for the max concurrent jobs.

JohnRoesler (approved):
I’m good with it. I’ll probably add some additional documentation on the risks of running the distributed locker with a max concurrent limit set.

git-hulk:
Cool, thanks for your help.

git-hulk:
@JohnRoesler Looks it has the data race in test case, will investigate and fix it.

JohnRoesler:
👍 thank you

PR #473, opened 2023-05-06
fix: deadlock for blocking channel

rfyiamcool:
### What does this do?

This is a scenario that occurs with a very small probability. The `StartBlocking()` and `Stop()` methods are executed almost concurrently. When the stop is notified through the channel, the start has not been monitored, so a deadlock problem occurs.

When will the stop method and start be executed concurrently, such as closing gocron after receiving a linux signal.

image

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -1377,7 +1377,7 @@ func (s *Scheduler) CustomTimer(customTimer func(d time.Duration, f func()) *tim
 func (s *Scheduler) StopBlockingChan() {
 	s.startBlockingStopChanMutex.Lock()
 	if s.startBlockingStopChan != nil {
-		s.startBlockingStopChan <- struct{}{}
+		close(s.startBlockingStopChan)
 	}
 	s.startBlockingStopChanMutex.Unlock()
 }
--- a/scheduler.go
+++ b/scheduler.go
@@ -82,7 +82,12 @@ func (s *Scheduler) StartBlocking() {
 	s.startBlockingStopChanMutex.Lock()
 	s.startBlockingStopChan = make(chan struct{}, 1)
 	s.startBlockingStopChanMutex.Unlock()
+
 	<-s.startBlockingStopChan
+
+	s.startBlockingStopChanMutex.Lock()
+	s.startBlockingStopChan = nil
+	s.startBlockingStopChanMutex.Unlock()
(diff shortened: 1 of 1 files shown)


scheduler.go
```diff
@@ -1377,7 +1377,7 @@ func (s *Scheduler) CustomTimer(customTimer func(d time.Duration, f func()) *tim
 func (s *Scheduler) StopBlockingChan() {
 	s.startBlockingStopChanMutex.Lock()
 	if s.startBlockingStopChan != nil {
-		s.startBlockingStopChan <- struct{}{}
+		close(s.startBlockingStopChan)
```
JohnRoesler, line 1380: this can result in a panic if the method is called twice.

If you modify the stop method to look like:
```golang
func (s *Scheduler) stop() {
	s.stopJobs(s.jobs)
	s.executor.stop()
	s.StopBlockingChan()
	s.setRunning(false)
}
```
and this method to look like 

```golang
func (s *Scheduler) StopBlockingChan() {
	s.startBlockingStopChanMutex.Lock()
	if s.IsRunning() && s.startBlockingStopChan != nil {
		close(s.startBlockingStopChan)
	}
	s.startBlockingStopChanMutex.Unlock()
}
```
i believe that solves it

JohnRoesler: I also added setting the channel to nil after closing so that if the method is called when someone does start async it won’t also panic on a closed channel

JohnRoesler: Hm, looks like that resulted in a data race in the tests. I’ll look later if you’re aren’t able to get to it.

JohnRoesler: @rfyiamcool this is the change I was looking to see for the StartBlocking method

```diff
func (s *Scheduler) StartBlocking() {
	s.StartAsync()
	s.startBlockingStopChanMutex.Lock()
	s.startBlockingStopChan = make(chan struct{}, 1)
	s.startBlockingStopChanMutex.Unlock()
	<-s.startBlockingStopChan
+	s.startBlockingStopChanMutex.Lock()
+	s.startBlockingStopChan = nil
+	s.startBlockingStopChanMutex.Unlock()
}
```

JohnRoesler (approved):
thank you!

PR #474, opened 2023-05-07
Fix data race in executor with enabling the limit mode

git-hulk:
### What does this do?

The root cause is we write the limitModeFuncsRunning(a wait group) in the new goroutine, so the Go data race detector will think they have the data race due to reading and writing in different routines.

I'm sorry for didn't run tests in data race mode in the previous PR. @JohnRoesler 

Fix the data race in https://github.com/go-co-op/gocron/actions/runs/4905158223/jobs/8758756592

None

None

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -96,7 +96,6 @@ func (jf *jobFunction) singletonRunner() {
 }
 
 func (e *executor) limitModeRunner() {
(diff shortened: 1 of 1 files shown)


JohnRoesler (approved):

JohnRoesler:
Thank you

PR #475, opened 2023-05-08
fix: executor run block when executor.stop()

rfyiamcool:
### What does this do?

When we close gocron, the `limitModeRunner` goroutine exits after receive ctx.Done().  Because there is no method to listen `limitModeQueue`, will pushing data to the `limitModeQueue` be blocked ？

`this is my guess`

- [ ] Updated `README.md`


JohnRoesler (approved):
always good to be safe than sorry 👍

PR #478, opened 2023-05-09
Add support of creating the lock key for the job when the distributed lock was enabled

### LINKED ISSUE #472: [FEATURE] Add support of the alias name for job
Currently, the executor prevents the same job from being run more than once by the function name, and need to create many different functions if we wanna create a cron job per user. So I'm wondering if it's possible to allow users to set the alias name for the job.

Add a new API like `DoWithName()` to create a job with the customized name, and use this name as the distributed locking key.

I will submit the PR if this proposal sounds good to you. @JohnRoesler

git-hulk:

This closes #472 

None

Yes

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -132,11 +132,11 @@ func (e *executor) start() {
 func (e *executor) runJob(f jobFunction) {
 	switch f.runConfig.mode {
 	case defaultMode:
-		lockKey := f.lockKey
-		// Use the function name as the lock key if no lock key is provided
+		lockKey := f.jobName
+		// Use the function funcName as the lock key if no lock key is provided
 		// via the Scheduler.SetJobLockKey() method
-		if f.lockKey == "" {
-			lockKey = f.name
+		if f.jobName == "" {
+			lockKey = f.funcName
 		}
 		if e.distributedLocker != nil {
 			l, err := e.distributedLocker.Lock(f.ctx, lockKey)
@@ -196,7 +196,7 @@ func (e *executor) run() {
 				if panicHandler != nil {
 					defer func() {
 						if r := recover(); r != any(nil) {
-							panicHandler(f.name, r)
+							panicHandler(f.funcName, r)
 						}
 					}()
 				}
--- a/executor.go
+++ b/executor.go
@@ -133,9 +133,7 @@ func (e *executor) runJob(f jobFunction) {
 	switch f.runConfig.mode {
 	case defaultMode:
 		lockKey := f.jobName
-		// Use the function funcName as the lock key if no lock key is provided
-		// via the Scheduler.SetJobLockKey() method
-		if f.jobName == "" {
+		if lockKey == "" {
 			lockKey = f.funcName
 		}
 		if e.distributedLocker != nil {
--- a/executor.go
+++ b/executor.go
@@ -132,8 +132,14 @@ func (e *executor) start() {
 func (e *executor) runJob(f jobFunction) {
 	switch f.runConfig.mode {
 	case defaultMode:
+		lockKey := f.lockKey
+		// Use the function name as the lock key if no lock key is provided
+		// via the Scheduler.SetJobLockKey() method
+		if f.lockKey == "" {
+			lockKey = f.name
+		}
 		if e.distributedLocker != nil {
-			l, err := e.distributedLocker.Lock(f.ctx, f.name)
+			l, err := e.distributedLocker.Lock(f.ctx, lockKey)
 			if err != nil || l == nil {
 				return
 			}
--- a/job.go
+++ b/job.go
@@ -45,8 +45,8 @@ type jobFunction struct {
 	function          any                // task's function
 	parameters        []any              // task's function parameters
 	parametersLen     int                // length of the passed parameters
-	lockKey           string             // key of the distributed lock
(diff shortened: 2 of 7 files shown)


git-hulk:
Hi @jasonlvhit

I finally didn't add the job name for the Job struct since we're embedding the jobFunction and it also has a name field. So it will be confusing and error-prone to add a similar name.

job.go
```diff
@@ -45,6 +45,7 @@ type jobFunction struct {
 	function          any                // task's function
 	parameters        []any              // task's function parameters
 	parametersLen     int                // length of the passed parameters
+	lockKey           string             // key of the distributed lock
 	name              string             // nolint the function name to run
```
JohnRoesler, line 49: i'd like to change the naming a bit here so this isn't scoped to only distributed locking
```suggestion
	jobName           string             // key of the distributed lock
	funcName          string             // the name of the function - e.g. main.func1
```

scheduler.go
```diff
@@ -126,6 +126,14 @@ func (s *Scheduler) Jobs() []*Job {
 	return s.jobs
 }
 
+// SetJobLockKey sets the lock key of the current job to prevent running jobs
+// more than once between multiple instances when the distributed lock is enabled.
+func (s *Scheduler) SetJobLockKey(name string) *Scheduler {
+	job := s.getCurrentJob()
+	job.lockKey = name
```
JohnRoesler, line 133:
```suggestion
// Name sets the name of the current job. 
//
// If the scheduler is running using WithDistributedLocker(), the job name is used
// as the distributed lock key.
func (s *Scheduler) Name(name string) *Scheduler {
	job := s.getCurrentJob()
	job.jobName = name
```

JohnRoesler: the job should also expose this name method

```golang
func (j *Job) Name(name string) {
	j.mu.Lock()
	defer j.mu.Unlock()
	j.jobName = name
}
```

git-hulk:
@JohnRoesler Thanks for your quick review, applied all suggestions in the last commit.

executor.go
```diff
@@ -132,8 +132,14 @@ func (e *executor) start() {
 func (e *executor) runJob(f jobFunction) {
 	switch f.runConfig.mode {
 	case defaultMode:
+		lockKey := f.jobName
+		// Use the function funcName as the lock key if no lock key is provided
+		// via the Scheduler.SetJobLockKey() method
+		if f.jobName == "" {
+			lockKey = f.funcName
+		}
```
JohnRoesler, line 140:
```suggestion
		lockKey := f.jobName
		if lockKey == "" {
			lockKey = f.funcName
		}
```

JohnRoesler:
thanks for all your work contributing 😄

scheduler_test.go
```diff
@@ -2625,3 +2625,61 @@ func runTestWithDistributedLocking(t *testing.T, maxConcurrentJobs int) {
...
+		schedulers[i].Stop()
+	}
+	close(resultChan)
+
+	var results []int
+	for r := range resultChan {
+		results = append(results, r)
+	}
+	assert.Len(t, results, expectedResult)
+}
```
JohnRoesler, line 2685: I prefer table driven testing, which is essentially what you're doing 😄 just a little tweak https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
```suggestion
func TestScheduler_WithDistributedLocker_With_Name(t *testing.T) {
	testCases := []struct {
		description string
		name        string
		expected    int
	}{
		{"with job name set", "foo", 4},
		// The distributed locking will use the function funcName as its lock key
		// if the job funcName is empty. So the result will be 8 instead of 4 since
		// they are regarded as different jobs.
		{"without job name", "", 8},
	}

	for _, tc := range testCases {
		t.Run(tc.description, func(t *testing.T) {
			resultChan := make(chan int, 10)
			f1 := func(schedulerInstance int) {
				resultChan <- schedulerInstance
				time.Sleep(100 * time.Millisecond)
			}
			f2 := func(schedulerInstance int) {
				resultChan <- schedulerInstance
				time.Sleep(100 * time.Millisecond)
			}

			l := &locker{
				store: make(map[string]struct{}, 0),
			}

			schedulers := make([]*Scheduler, 0)

			var err error
			var job *Job
			for i := 0; i < 2; i++ {
				s := NewScheduler(time.UTC)
				s.WithDistributedLocker(l)
				if i%2 == 0 {
					job, err = s.Every("500ms").Name(tc.name).Do(f1, 1)
				} else {
					job, err = s.Every("500ms").Name(tc.name).Do(f2, 1)
				}
				require.NoError(t, err)
				require.Equal(t, job.jobName, tc.name)
				schedulers = append(schedulers, s)
			}
			for i := range schedulers {
				schedulers[i].StartAsync()
			}
			time.Sleep(1700 * time.Millisecond)
			for i := range schedulers {
				schedulers[i].Stop()
			}
			close(resultChan)

			var results []int
			for r := range resultChan {
				results = append(results, r)
			}
			assert.Len(t, results, tc.expected)
		})
	}
}
```

JohnRoesler: @git-hulk just want to make sure you saw this 😄

JohnRoesler (approved):

PR #479, opened 2023-05-09
fix: make panic on startblocking with no jobs more clear

JohnRoesler:
### What does this do?
as the comment in the code notes: 

```
// panic here because if we don't, a panic will result due to a deadlock
// as all the goroutines will be asleep as no jobs are being published
// to the executor from the scheduler. This way, the panic message will
// be clearer to the end user.
```

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -77,11 +77,23 @@ func (s *Scheduler) SetMaxConcurrentJobs(n int, mode limitMode) {
 
 // StartBlocking starts all jobs and blocks the current thread.
 // This blocking method can be stopped with Stop() from a separate goroutine.
+//
+// NOTE - This will panic if called with zero jobs in the scheduler.
 func (s *Scheduler) StartBlocking() {
 	s.StartAsync()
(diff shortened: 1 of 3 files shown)


scheduler.go
```diff
@@ -77,11 +77,23 @@ func (s *Scheduler) SetMaxConcurrentJobs(n int, mode limitMode) {
...
 func (s *Scheduler) StartBlocking() {
 	s.StartAsync()
 	s.startBlockingStopChanMutex.Lock()
 	s.startBlockingStopChan = make(chan struct{}, 1)
 	s.startBlockingStopChanMutex.Unlock()
+	if len(s.Jobs()) == 0 {
+		time.Sleep(time.Second)
+		if len(s.Jobs()) == 0 {
+			// panic here because if we don't, a panic will result due to a deadlock
+			// as the all goroutines will be asleep as no jobs are being published
```
JohnRoesler, line 91:
```suggestion
			// as all the goroutines will be asleep as no jobs are being published
```

Issue #486, opened 2023-05-12
[BUG] - panic: runtime error: integer divide by zero


artking28:
### Describe the bug

My runtime is exiting by it, I dont know why, its only When I use ur library

My code 
```
func LaunchCronTasks(port string) error {
	s := gocron.NewScheduler(time.UTC)

	
	if _, err := s.Every(30).Seconds().StartAt(time.Now().Add(10 * time.Second)).Do(func() []error {
		var errs []error
		errs = append(errs, myFunction())
		return errs
	}); err != nil {
		return err
	}

	if s.Len() > 0 {
		s.StartAsync()
		println("\033[32m" + "[PORT" + port + "] Cron launched with success       --> " + utils.IntoString(time.Now()) + "\033[0m")
	}
	return nil
}
```

### Version

v1.18.0

![Captura de ecrã 2023-05-11, às 21 39 06](https://github.com/go-co-op/gocron/assets/78158057/7cf522f5-1ef6-4fd6-9248-30ce3420b37f)

JohnRoesler:
@artking28 have you tried the latest version?

artking28:
Thank u

BranislavLazic:
@JohnRoesler 1.28.0 version - still the same problem. Basically, this is the cause: `StartAt(time.Now().Add(10 * time.Second))`

PR #504, opened 2023-05-31
fix: panic: runtime error: integer divide by zero

BranislavLazic:
"Do" function triggers "doCommon" function that sets "inScheduleChain" to false. In "getCurrentJob", if "inScheduleChain" is false, then a new job will be added to a slice and the current job will be the last element of the slice. Therefore, after "Do" function runs, "StartAt" will try to obtain the current job. Since "inScheduleChain" is false, a new job with interval 0 will be added and returned as the current job in "StartAt" function. After this interval 0 will produce "duration" to be 0 and "panic: runtime error: integer divide by zero".

Resolves https://github.com/go-co-op/gocron/issues/486

Yes

- [ ] Updated `README.md`

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -756,7 +756,7 @@ func ExampleScheduler_StartAt() {
 	s.StartBlocking()
 }
 
-func ExampleScheduler_StartAt_AfterTaskStart() {
+func ExampleScheduler_StartAt_afterTaskStart() {
 	s := gocron.NewScheduler(time.UTC)
 	_, _ = s.Every(5).Seconds().Do(task)
 	s.StartAt(time.Now().Add(10 * time.Second))
--- a/example_test.go
+++ b/example_test.go
@@ -756,6 +756,13 @@ func ExampleScheduler_StartAt() {
 	s.StartBlocking()
 }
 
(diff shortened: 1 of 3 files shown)


example_test.go
```diff
@@ -756,6 +756,13 @@ func ExampleScheduler_StartAt() {
 	s.StartBlocking()
 }
 
+func ExampleScheduler_StartAt_afterTaskStart() {
+	s := gocron.NewScheduler(time.UTC)
+	_, _ = s.Every(5).Seconds().Do(task)
+	s.StartAt(time.Now().Add(10 * time.Second))
```
JohnRoesler, line 762: Hi @BranislavLazic thanks for taking a stab at this. The solution you have here actually changes / breaks the design of the scheduling chain. The `inScheduleChain` bool is to prevent this type of activity as it causes more pain confusion to users not knowing which job exactly is being changed.

The chain is started anytime a method is called on the scheduler and terminated with `Do` is called.

The actual issue where a divide by zero can happen is when startAt is called and the time is in the past. I think just wrapping the potential divide by zero case in a check if the denominator is zero first will be the way to solve it. I'll open a PR

PR #531, opened 2023-07-30
feat: handle pointer to func

ErfanMomeniii:
Hi,
Merging this pull request can add the ability to pass the pointer of func in `Do()` method.
Example:
```go
fn := func() { fmt.Println("Hi") }
_, err := s.Every(1).Second().Do(&fn)
fmt.Println(err)
// nil

_, err = s.Every(1).Second().Do(fn)
fmt.Println(err)
// nil
```
I have added a test for that, and it works properly and is also backward compatible.
If you have any ideas or concerns regarding the PR, feel free to tell me.

### DIFF
--- a/gocron.go
+++ b/gocron.go
@@ -127,6 +127,10 @@ func getFunctionName(fn interface{}) string {
 	return runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 }
 
+func getFunctionNameOfPointer(fn interface{}) string {
+	return runtime.FuncForPC(reflect.ValueOf(fn).Elem().Pointer()).Name()
+}
+
 func parseTime(t string) (hour, min, sec int, err error) {
 	var timeLayout string
 	switch {
--- a/scheduler.go
+++ b/scheduler.go
@@ -932,21 +932,37 @@ func (s *Scheduler) doCommon(jobFun interface{}, params ...interface{}) (*Job, e
 		return nil, job.error
 	}
 
-	typ := reflect.TypeOf(jobFun)
-	if typ.Kind() != reflect.Func {
+	val := reflect.ValueOf(jobFun)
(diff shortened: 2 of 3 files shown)


gocron.go
```diff
@@ -127,6 +127,10 @@ func getFunctionName(fn interface{}) string {
 	return runtime.FuncForPC(reflect.ValueOf(fn).Pointer()).Name()
 }
 
+func getFunctionNameOfPointer(fn interface{}) string {
```
JohnRoesler, line 130: hey @ErfanMomeniii thanks for the thorough contribution! 

Could you help me understand what the use-case is for passing a pointer to a function into the scheduler?

ErfanMomeniii:
Hi, dear @JohnRoesler
I used this package in another project, and this thing caused an error so I went to add this feature to support pointer to func.
I believe it might be beneficial to add this.

scheduler.go
```diff
@@ -932,21 +932,39 @@ func (s *Scheduler) doCommon(jobFun interface{}, params ...interface{}) (*Job, e
...
+	if val == reflect.ValueOf(jobFun) {
+		fname = getFunctionName(jobFun)
+	} else {
+		fname = getFunctionNameOfPointer(jobFun)
+	}
+
 	if job.funcName != fname {
 		job.function = jobFun
+		if val != reflect.ValueOf(jobFun) {
+			job.function = reflect.ValueOf(jobFun).Elem().Interface()
```
JohnRoesler, line 956: val would already be the reflect valueOf and the Elem if it's a pointer
```suggestion
			job.function = val.Interface()
```

ErfanMomeniii:
all problems should be fixed now.

JohnRoesler (approved):
thanks @ErfanMomeniii

PR #559, opened 2023-08-31
fix: block unlock with limit concurrency

rfyiamcool:
### change

see title 😁


JohnRoesler:
Do you have sample code or a test that creates the issue?

JohnRoesler (approved):
thanks for the test 👍

PR #567, opened 2023-09-18
Fix some data races

knightpp:
### What does this do?
I noticed data races in my codebase, turns out the issue is not in my code. So, I fixed it here.

Tests pass

No

- [ ] Updated `README.md`

### Notes
I've not tested it, only run `go test ./...`. Feel free to close if you feel that this PR is not appropriate.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -95,15 +95,15 @@ func (s *Scheduler) StartBlocking() {
 
 // StartAsync starts all jobs without blocking the current thread
 func (s *Scheduler) StartAsync() {
-	if !s.IsRunning() {
+	if s.running.CAS(false, true) {
 		s.start()
 	}
 }
 
 // start starts the scheduler, scheduling and running jobs
 func (s *Scheduler) start() {
-	s.executor.start()
 	s.setRunning(true)
+	s.executor.start()
 	s.runJobs(s.jobsMap())
 }
 
--- a/scheduler.go
+++ b/scheduler.go
@@ -95,15 +95,15 @@ func (s *Scheduler) StartBlocking() {
 
 // StartAsync starts all jobs without blocking the current thread
 func (s *Scheduler) StartAsync() {
-	if s.running.CAS(false, true) {
+	if !s.IsRunning() {
 		s.start()
 	}
 }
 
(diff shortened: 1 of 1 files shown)


scheduler.go
```diff
@@ -95,19 +95,22 @@ func (s *Scheduler) StartBlocking() {
 
 // StartAsync starts all jobs without blocking the current thread
 func (s *Scheduler) StartAsync() {
-	if !s.IsRunning() {
+	if s.running.CAS(false, true) {
```
JohnRoesler, line 98: CAS is deprecated, docs say to use `CompareAndSwap`

JohnRoesler: But, I wouldn't use that here. I'd leave it as `!s.IsRunning()`

JohnRoesler:
Thanks for the contribution! 

Could you share the data races you saw? The tests are failing on this PR because of new data races that weren't previously there, so this doesn't seem to be solving the problem.

scheduler.go
knightpp: I didn't notice because go.mod depends on `go.uber.org/atomic v1.9.0` which does not yet have `CompareAndSwap`. 

There's an issue in `if !s.IsRunnin()`. Someone can start multiple goroutines calling this function and will probably get a race condition because it might call s.start() multiple times.

knightpp:
I cannot provide code with the issue, but the data race happens when [getCurrentJob()](https://github.com/go-co-op/gocron/blob/7fea08c327f0ac9daa73966c9f4233f2b0102685/scheduler.go#L1317) sets an element in the map and [runJobs()](https://github.com/go-co-op/gocron/blob/7fea08c327f0ac9daa73966c9f4233f2b0102685/scheduler.go#L111) tries to iterate over the map.

scheduler.go
knightpp: It feels like settings s.running to true too early results in a start of multiple executors, so reverted that change.

knightpp:
Also, the fix should probably be applied to every place where `jobsMap()` is used in for range loop.

JohnRoesler:
#589 solved this issue

PR #575, opened 2023-09-21
fix Scheduler.NextRun() not returning the correct value (#574)

### LINKED ISSUE #574: [BUG] - Scheduler.NextRun() always returns 0001-01-01 00:00:00 +0000 UTC
### To Reproduce

Example
```golang
s2 := gocron.NewScheduler(time.UTC)
s2.StartAsync()
s2.Every(1).Friday().At("20:19").Do(func() {})
fmt.Println(s2.NextRun())
```
The output :
```shell
<nil> 0001-01-01 00:00:00 +0000 UTC
```
### Version

v1.34.1

### Expected behavior

Return to the correct Job and time

### Additional context

https://github.com/go-co-op/gocron/blob/01fcdee22fd5bf22f1be6b7dab52a32647a0bd1f/scheduler.go#L522-L531

The variable `nearestRun` is initialized to `0001-01-01 00:00:00 +0000 UTC`
It looks like `nr.Before(nearestRun)` will always be false

bbaa-bbaa:
### What does this do?
Initialize nearestRun to the next execution time of the first Job in map.

Resolves #574

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -520,8 +520,15 @@ func (s *Scheduler) NextRun() (*Job, time.Time) {
 
 	var jobID uuid.UUID
 	var nearestRun time.Time
+	flag := false
 	for _, job := range s.jobsMap() {
 		nr := job.NextRun()
+		if !flag && s.now().Before(nr) {
+			nearestRun = nr
+			jobID = job.id
+			flag = true
+			continue
+		}
 		if nr.Before(nearestRun) && s.now().Before(nr) {
 			nearestRun = nr
 			jobID = job.id


scheduler.go
```diff
@@ -520,8 +520,15 @@ func (s *Scheduler) NextRun() (*Job, time.Time) {
...
+	flag := false
 	for _, job := range s.jobsMap() {
 		nr := job.NextRun()
+		if !flag && s.now().Before(nr) {
+			nearestRun = nr
+			jobID = job.id
+			flag = true
+			continue
+		}
 		if nr.Before(nearestRun) && s.now().Before(nr) {
```
JohnRoesler, line 532: we can simplify the check here a little 😄 
```suggestion
	for _, job := range s.jobsMap() {
		nr := job.NextRun()
		if (nr.Before(nearestRun) || nearestRun.IsZero()) && s.now().Before(nr) {
```

JohnRoesler:
@bbaa-bbaa thank you so much for opening an issue and making a PR to solve it 😄

JohnRoesler (approved):

PR #603, opened 2023-10-30
fix(scheduler): ensure negative intervals given to `Every` return an immediate error #600

### LINKED ISSUE #600: [BUG] - Handling negative time.Duration as interval param to `Every` func
### Describe the bug

Not sure if this is a bug, more of a question to confirm expected behavior: currently no error is returned when a negative `time.Duration` is provided as the interval param to `Every`. Is that intentional? Should it not return an error and fail fast, similar to when providing a negative int? Not clear to me what the intended behavior is for that.

### To Reproduce

```
scheduler := gocron.NewScheduler(time.UTC)
interval := -1 * time.Second
_, err := scheduler.Every(interval).Do(func() { })
```

### Version

v1.35.2

### Expected behavior

`err` is returned non-nil, or otherwise behavior for negative Duration is documented on the `Every` function.

### Additional context

Thinking it would be helpful as input validation to get an error in this scenario to immediately recognize invalid setup, just like how negative int values are treated as errors.

husam-e:
### What does this do?

Ensures negative intervals return an immediate error.

Resolves #600 

Negative intervals will now return an error.

Yes.

- [ ] Updated `README.md`

### Notes

Potentially breaking change for any consumer expecting negative intervals to not return an error.

### DIFF
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -145,6 +145,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 6, counter)
+
+		_, err = s.Every(-1 * time.Millisecond).Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval)
 	})
 
 	t.Run("int", func(t *testing.T) {
@@ -192,6 +198,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 2, counter)
+
+		_, err = s.Every("-1ms").Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval)
 	})
 }
 
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -150,7 +150,7 @@ func TestScheduler_Every(t *testing.T) {
 			// do nothing
 		})
 
-		require.EqualError(t, err, ErrInvalidInterval.Error())
+		require.ErrorIs(t, err, ErrInvalidInterval)
 	})
 
 	t.Run("int", func(t *testing.T) {
@@ -203,7 +203,7 @@ func TestScheduler_Every(t *testing.T) {
 			// do nothing
 		})
 
-		require.EqualError(t, err, ErrInvalidInterval.Error())
+		require.ErrorIs(t, err, ErrInvalidInterval)
 	})
 }
 
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -150,7 +150,7 @@ func TestScheduler_Every(t *testing.T) {
 			// do nothing
 		})
 
-		require.ErrorIs(t, err, ErrInvalidInterval)
+		require.Contains(t, err.Error(), ErrInvalidInterval.Error())
 	})
 
 	t.Run("int", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


scheduler_test.go
```diff
@@ -145,6 +145,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 6, counter)
+
+		_, err = s.Every(-1 * time.Millisecond).Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval)
```
JohnRoesler: > cannot use ErrInvalidInterval (variable of type error) as string value in argument to require.EqualError
```suggestion
		require.EqualError(t, err, ErrInvalidInterval.Error())
```

scheduler_test.go
```diff
@@ -192,6 +198,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 2, counter)
+
+		_, err = s.Every("-1ms").Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval)
```
husam-e:
```suggestion
		require.EqualError(t, err, ErrInvalidInterval.Error())
```

scheduler_test.go
```diff
@@ -145,6 +145,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 6, counter)
+
+		_, err = s.Every(-1 * time.Millisecond).Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval.Error())
```
JohnRoesler: ah, the error is wrapped. So it's got another one around it. I think using `require.ErrorIs` should do it

```suggestion
		require.ErrorIs(t, err, ErrInvalidInterval)
```

scheduler_test.go
```diff
@@ -192,6 +198,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 2, counter)
+
+		_, err = s.Every("-1ms").Do(func() {
+			// do nothing
+		})
+
+		require.EqualError(t, err, ErrInvalidInterval.Error())
```
```suggestion
		require.ErrorIs(t, err, ErrInvalidInterval)
```

scheduler_test.go
```diff
@@ -145,6 +145,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 6, counter)
+
+		_, err = s.Every(-1 * time.Millisecond).Do(func() {
+			// do nothing
+		})
+
+		require.ErrorIs(t, err, ErrInvalidInterval)
```
```suggestion
		require.Contains(t, err.Error(), ErrInvalidInterval.Error())
```

scheduler_test.go
```diff
@@ -192,6 +198,12 @@ func TestScheduler_Every(t *testing.T) {
 		}
 		s.Stop()
 		assert.Equal(t, 2, counter)
+
+		_, err = s.Every("-1ms").Do(func() {
+			// do nothing
+		})
+
+		require.ErrorIs(t, err, ErrInvalidInterval)
```
```suggestion
		require.Contains(t, err.Error(), ErrInvalidInterval.Error())
```

JohnRoesler (approved):

scheduler_test.go
husam-e: thanks, I actually removed these checks as I realized there was already a bad interval test case, so I added cases to that and switched it to use contain.

PR #606, opened 2023-11-09
Exposing job's  scheduled unit and interval

chicknsoup:
### What does this do?
Exposes job's scheduled unit and interval
Added String() function to schedulingUnit 

None

No need

Yes

### DIFF
--- a/job.go
+++ b/job.go
@@ -447,8 +447,8 @@ func (j *Job) ScheduledTime() time.Time {
 }
 
 // ScheduledUnit returns the scheduled unit of the Job.
-func (j *Job) ScheduledUnit() schedulingUnit {
-        return j.unit
+func (j *Job) ScheduledUnit() string {
+        return j.unit.String()
 }
 
 // Interval returns the scheduled interval of the Job.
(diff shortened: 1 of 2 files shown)


job.go
```diff
@@ -446,6 +446,16 @@ func (j *Job) ScheduledTime() time.Time {
 	return j.nextRun
 }
 
+// ScheduledUnit returns the scheduled unit of the Job.
+func (j *Job) ScheduledUnit() schedulingUnit {
```
JohnRoesler: there should be a public version of schedulingUnit - or perhaps just make `schedulingUnit` and all the const values public

job.go
```diff
@@ -446,6 +446,16 @@ func (j *Job) ScheduledTime() time.Time {
...
+        return j.unit
```
JohnRoesler: please run `gofumpt -w .`

chicknsoup: Done, thank you

job.go
chicknsoup: Changed the return type to string. I think that should be enough.

JohnRoesler (approved):

PR #607, opened 2023-11-09
add job storage

### LINKED ISSUE #533: [FEATURE] - Can gocron read from a persistent storage the scheduled jobs and schedule
### Is your feature request related to a problem? 

We are running gocron in a containerized environment where jobs are scheduled based on rest api inputs. 
We would like to use some sort of persistence to store the received jobs details like frequency etc.
When the container restarts / the instance where container is running restarts we would like the gocron scheduler to fetch
the existing job status and schedule the next run of each job accordingly.

Solution would be that on a restart , the scheduler will be able load data from persistent storage and continue the subsequent runs as it would have  if system hadnt rebooted/restarted. The scheduler would update the last run time in the persistent storage so that on fresh startup, it can recalculate the next run time.

Have a wrapper function, which would be responsible for updating persistent storage on when the run was last scheduled. On fresh restart , it would load the scheduler by recalibrating the start time from the last run time and add the jobs.

pkoukk:
### What does this do?
It allows the scheduler to load jobs from a job object, which may come from persistent storage.

And it reports job run results to a reporter interface, which can be helpful for job resuming.

Resolves #533 

No break changes.

Yes. 

1.  `TestLoadJobFromObject` in `scheduler_test.go`
2. `Test_ExecutorJobReport` in `executor_test.go`

- [ ] Updated `README.md`

### DIFF
--- a/storage.go
+++ b/storage.go
@@ -1,0 +1,40 @@
+package gocron
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type JobObject struct {
+	UUID    uuid.UUID
+	JobName string
+	LastRun time.Time
+
+	Cron            string
+	CronWithSeconds bool
+
+	Every interface{}
+
+	Unit    string
+	WeekDay *time.Weekday
+	Months  []int
+
+	At []interface{}
+}
+
+type JobStorage interface {
+	LoadJobs() ([]JobObject, error)
+}
+
+type JobResult struct {
+	UUID       uuid.UUID
+	JobName    string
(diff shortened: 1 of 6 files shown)


storage.go
```diff
@@ -0,0 +1,40 @@
+package gocron
+
+import (
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type JobObject struct {
```
JohnRoesler, line 9: what about the actual job function? Right now, only the job run information is stored and not the actual function to be run 🤔

PR #613, opened 2023-11-13
force `executor` to use attached logger instead of "log" package

AlphaNecron:
### What does this do?
This forces `executor` to use logger instance instead of `log` package to allow customizability and ensure consistency among log messages.

Nothing

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -2,7 +2,6 @@ package gocron
 
 import (
 	"context"
-	"log"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


JohnRoesler (approved):
Thanks for catching that!

AlphaNecron:
My fault, that should be `e.logger.Debug`, not `e.logger.Info` anyways.

AlphaNecron:
My bad, that should be `e.logger.Debug`, not `e.logger.Info`.

JohnRoesler:
Ah, and my bad for not catching in review. Not a big deal

Issue #624, opened 2023-11-21
[ENHANCEMENT] - SetMaxConcurrentJobs to work with singleton jobs


hspark0582:
### Describe the bug

Singleton jobs do not follow the limit set with `SetMaxConcurrentJobs`.

### To Reproduce

Test code:
```go
// skipping headers

func main() {
	s := gocron.NewScheduler(time.UTC)

	s.SetMaxConcurrentJobs(2, gocron.WaitMode)
	s.SingletonModeAll()

	for i := 0; i < 4; i++ {
		s.Every(time.Minute).Do(
			func() {
				logrus.Infof("Run some job")
				exec.Command("sleep", "5").Run()
			},
		)
	}

	s.StartBlocking()
}
```
(Error handling omitted for brevity)

Output shows 4 "Run some job"s immediately, ignoring `SetMaxConcurrentJobs` of 2.
When `s.SingletonModeAll()` was removed, only 2 "Run some job"s are displayed as intended.

### Version

v1.36.0

### Expected behavior

`SetMaxConcurrentJobs` also apply to singleton jobs.
The number of jobs, normal and singleton jobs combined, does not exceed the limit set with `SetMaxConcurrentJobs`.

JohnRoesler:
hello @hspark0582, could you help me understand the expected behavior and why you would combine singleton and limit mode together?

hspark0582:
Hello!

I was using gocron to schedule jobs that are somewhat heavy in CPU load, thus I wanted to limit the number of concurrent jobs. Also, those jobs may crash when multiple jobs run simultaneously, so I made them all singleton. Not much to say here really.

I actually have an update, just to provide a bit of my usage: I "solved" this issue with making jobs regular (not singleton) and `if job.RunCount() - job.FinishedRunCount() > 1 { return }`.

Expected behavior is, well, *my* expected behavior, so I wouldn't mind you disagreeing with me.

I don't know if this helped, so if you need more info, please ask. Thank you all for maintaining this wonderful work.

JohnRoesler:
@hspark0582 thanks for the additional context. That makes sense. It would take some thinking about, as far as how to handle it. The current design in v1 & v2 assumes them to be mutually exclusive. v1 gives precedence to singleton mode, whereas v2 gives precedence to limit mode. 

I can see how you'd want to make sure that if you have a limit of 2, that you don't end up with the same job running twice and taking both the limit spots.

JohnRoesler:
[...]

PR #625, opened 2023-11-28
allow max concurrent runs and singleton mode together

JohnRoesler:
### What does this do?
allows for singleton mode and max concurrent runs to be combined. The resulting behavior is that if you have a limit of say 2 jobs concurrently, you can't have one job take both those slots. By having singleton enabled on a job, it is prevented from running by two limitModeRunners simultaneously. 

#624 

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -36,10 +36,16 @@ type limitModeConfig struct {
 	limit             uint
 	rescheduleLimiter chan struct{}
 	in                chan uuid.UUID
+	// singletonJobs is used to track singleton jobs that are running
+	// in the limit mode runner. This is used to prevent the same job
+	// from running multiple times across limit mode runners when both
+	// a limit mode and singleton mode are enabled.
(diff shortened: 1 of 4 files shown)


executor.go
```diff
@@ -204,106 +200,85 @@ func (e *executor) start() {
...
+			ctx, cancel := context.WithCancel(e.ctx)
+			j := requestJobCtx(ctx, id, e.jobOutRequest)
+			cancel()
+			if j != nil {
+				if j.singletonMode {
+					e.limitMode.singletonJobsMu.Lock()
+					_, ok := e.limitMode.singletonJobs[id]
+					if ok {
+						// this job is already running, so don't run it
+						// but instead reschedule it
```
JohnRoesler, line 225: Effectively, singleton mode is always reschedule with max concurrent. Is there a need to allow for singleton jobs to wait? It seems like no

PR #632, opened 2023-12-12
Ability to register for events associated with jobs starting immediately

drwpls:
### What does this do?

This commit add some functions to support register events for job which is on creating and run immediately.

Job run immediately doesn't have before/after/onErr/onNoErr events. 

No 

Yes, it's in scheduler_test.go

Yes, in function header

### Notes
Consider to change RegisterEventListeners to change only current job

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -1527,6 +1527,46 @@ func (s *Scheduler) RegisterEventListeners(eventListeners ...EventListener) {
 	}
 }
 
+// BeforeJobRuns registers an event listener that is called before a job runs.
+func (s *Scheduler) BeforeJobRuns(eventListenerFunc func(jobName string)) *Scheduler {
+	job := s.getCurrentJob()
+	job.mu.Lock()
+	defer job.mu.Unlock()
+	job.eventListeners.beforeJobRuns = eventListenerFunc
+
+	return s
+}
+
+// AfterJobRuns registers an event listener that is called after a job runs.
+func (s *Scheduler) AfterJobRuns(eventListenerFunc func(jobName string)) *Scheduler {
+	job := s.getCurrentJob()
+	job.mu.Lock()
+	defer job.mu.Unlock()
+	job.eventListeners.afterJobRuns = eventListenerFunc
+
+	return s
+}
+
+// WhenJobStarts registers an event listener that is called when a job starts.
+func (s *Scheduler) WhenJobReturnsError(eventListenerFunc func(jobName string, err error)) *Scheduler {
+	job := s.getCurrentJob()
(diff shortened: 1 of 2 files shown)


JohnRoesler (approved):
This is a handy addition. Although, nothing in the current code prevents you from having event listeners on a job that starts immediately.

```golang
package main

import (
	"log"
	"time"

	"github.com/go-co-op/gocron"
)

func main() {
	j, err := s.Every(1).Seconds().StartImmediately().Do(func() {})
	if err != nil {
		log.Fatal(err)
	}
	j.RegisterEventListeners(
		gocron.BeforeJobRuns(func(jobName string) {
			// do something
		}),
	)
	s.StartBlocking()
}
```

drwpls:
> This is a handy addition. Although, nothing in the current code prevents you from having event listeners on a job that starts immediately.
> 
> ```go
> package main
> 
> import (
> 	"log"
> 	"time"
> 
> 	"github.com/go-co-op/gocron"
> )
> 
> func main() {
> 	s := gocron.NewScheduler(time.UTC)
> 	j, err := s.Every(1).Seconds().StartImmediately().Do(func() {})
> 	if err != nil {
> 		log.Fatal(err)
> 	}
> 	j.RegisterEventListeners(
> 		gocron.BeforeJobRuns(func(jobName string) {
> 			// do something
> 		}),
> 	)
> 	s.StartBlocking()
> }
> ```

Your code will only work if the startBlocking statement runs after setting up the jobs, and it can't register events for individual jobs.

One more question: why did you define eventListenerFunc  signature contain jobName (string) instead of *Job

JohnRoesler:
> Your code will only work if the startBlocking statement runs after setting up the jobs, and it can't register events for individual jobs.

Correct, that you have to setup all the jobs first and then start. However, that does set the event listeners for an individual job. It's being called on the Job `j`.

> why did you define eventListenerFunc signature contain jobName (string) instead of *Job

To avoid race conditions with the Job being a pointer. 

Are there other fields that would be useful in the event listeners? I can definitely see how more info could be useful / better. I'd be happy to consider enhancements in v2 👍

drwpls:
If you set up 2 jobs, both got event registered. I don't think that is for individual job.
https://github.com/go-co-op/gocron/blob/30921b708f1ef7910842cc63fc27fe383c7e5d81/scheduler.go#L1517C1-L1528C2

> Correct, that you have to setup all the jobs first and then start. However, that does set the event listeners for an individual job. It's being called on the Job `j`.

You already have mutex in job structure.
> To avoid race conditions with the Job being a pointer.

PR #655, opened 2024-01-10
Add metrics

### LINKED ISSUE #317: [FEATURE] - Add Prometheus Metrics For Job

gorodet-sky:
### What does this do?
Add interface for collect jobs execution metrics: counter, time elapsed

Resolves #317 

Yes, tests for counter was added

- [x] Updated `README.md`

### Notes
- It seems to me that you can add a test for elapsed time, but I don’t know how to do it nice
- In the future I want to make a metrics implementation for Prometheus in a separate repository

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -786,20 +786,20 @@ func ExampleWithTags() {
 }
 
 /*
-type exampleMonitorer struct {
+type exampleMonitor struct {
 	mu      sync.Mutex
 	counter map[string]int
 	time    map[string][]time.Duration
 }
 
-func newExampleMonitorer() *exampleMonitorer {
-	return &exampleMonitorer{
+func newExampleMonitor() *exampleMonitor {
+	return &exampleMonitor{
 		counter: make(map[string]int),
 		time:    make(map[string][]time.Duration),
 	}
 }
 
-func (t *exampleMonitorer) Inc(_ uuid.UUID, name string, _ []string, _ JobStatus) {
+func (t *exampleMonitor) Inc(_ uuid.UUID, name string, _ []string, _ JobStatus) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	_, ok := t.counter[name]
@@ -809,7 +809,7 @@ func (t *exampleMonitorer) Inc(_ uuid.UUID, name string, _ []string, _ JobStatus
 	t.counter[name]++
 }
 
-func (t *exampleMonitorer) WriteTiming(startTime, endTime time.Time, _ uuid.UUID, name string, _ []string) {
+func (t *exampleMonitor) WriteTiming(startTime, endTime time.Time, _ uuid.UUID, name string, _ []string) {
 	t.mu.Lock()
 	defer t.mu.Unlock()
 	_, ok := t.time[name]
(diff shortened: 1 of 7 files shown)


metrics.go
```diff
@@ -0,0 +1,23 @@
...
+// The different statuses of job that can be used.
+const (
+	Fail    JobStatus = "fail"
+	Success JobStatus = "success"
+)
+
+// Monitorer represents the interface to collect jobs metrics.
+type Monitorer interface {
+	Inc(id uuid.UUID, name string, status JobStatus)
+	WriteTiming(startTime, endTime time.Time, id uuid.UUID, name string)
```
JohnRoesler: I could see including job tags here being useful. Then within a specific metrics implementation you can make choices about how to use those, e.g. in prometheus, they would be labels like `name=foobar,tags=tag1,tag2,id=uuid,status=success`

executor.go
```diff
@@ -350,11 +351,21 @@ func (e *executor) runJob(j internalJob, shouldSendOut bool) {
...
+	}
 	if err != nil {
 		_ = callJobFuncWithParams(j.afterJobRunsWithError, j.id, j.name, err)
+		if e.monitorer != nil {
+			e.monitorer.Inc(j.id, j.name, Success)
+		}
 	} else {
 		_ = callJobFuncWithParams(j.afterJobRuns, j.id, j.name)
+		if e.monitorer != nil {
+			e.monitorer.Inc(j.id, j.name, Fail)
```
JohnRoesler:
```suggestion
			e.monitorer.Inc(j.id, j.name, Fail)
		}
	} else {
		_ = callJobFuncWithParams(j.afterJobRuns, j.id, j.name)
		if e.monitorer != nil {
			e.monitorer.Inc(j.id, j.name, Success)
```

JohnRoesler: These were backwards

example_test.go
```diff
@@ -784,3 +784,66 @@ func ExampleWithTags() {
...
+	)
+	s.Start()
+	time.Sleep(5 * time.Second)
+	_ = s.Shutdown()
+
+	fmt.Printf("Job %q total execute count: %d\n", name, monitorer.counter[name])
+	for i, val := range monitorer.time[name] {
+		fmt.Printf("Job %q execute #%d elapsed %.4f seconds\n", name, i+1, val.Seconds())
+	}
+}
```
JohnRoesler: probably best off mimicking how this is done for the Executor, and just put it inside the function but commented out. Right now, it won't render in godoc very well

executor.go
```diff
@@ -23,6 +23,7 @@ type executor struct {
 	limitMode        *limitModeConfig
 	elector          Elector
 	locker           Locker
+	monitorer        Monitorer
```
JohnRoesler:
```suggestion
	monitor        Monitor
```

metrics.go
```diff
@@ -0,0 +1,23 @@
...
+type JobStatus string
+
+// The different statuses of job that can be used.
+const (
+	Fail    JobStatus = "fail"
+	Success JobStatus = "success"
+)
+
+// Monitorer represents the interface to collect jobs metrics.
+type Monitorer interface {
```
JohnRoesler: Can you make this just `Monitor` throughout?

```suggestion
// Monitor represents the interface to collect jobs metrics.
type Monitor interface {
```

metrics.go
gorodet-sky: Done

executor.go
gorodet-sky: Fixed

example_test.go
gorodet-sky: Commented

metrics.go
gorodet-sky: Sure, done

PR #668, opened 2024-02-02
fix RunNow() when calling from a job returned by Jobs()

### LINKED ISSUE #667: BUG: gocron: Job: RunNow: scheduler unreachable
### Discussed in https://github.com/go-co-op/gocron/discussions/666

<div type='discussions-op-text'>

<sup>Originally posted by **kishenpatel** February  1, 2024</sup>
Hey team! I am trying to leverage the job .RunNow() command but even with the simplest implementation I seem to get the error "gocron: Job: RunNow: scheduler unreachable", please see example code below and please let me know if I might be doing something wrong?

```go
package main

import (
	"fmt"
	"time"

)

func main() {
	fmt.Println("Hello, 世界")

	sch, err := gocron.NewScheduler()
	if err != nil {
		fmt.Println("Failed to create new scheduler")
		fmt.Print(err)
		return
	}

	_, err = sch.NewJob(
		gocron.DurationJob(
			30*time.Second,
		),
		gocron.NewTask(
			func(a string, b int) {
				fmt.Println(a, b)
			},
			"hello",
			1,
		),
	)
	if err != nil {
		fmt.Println("Failed to add job")
		fmt.Print(err)
		return
	}
	fmt.Println("Successfully added job")

	sch.Start()

	jobs := sch.Jobs()
	j0 := jobs[0]
	err = j0.RunNow()
	if err != nil {
		fmt.Println("Failed to run job")
		fmt.Print(err)
	}
}

``` 

Playground:
https://go.dev/play/p/6MMUrYTeZDX

Please advise! Thank you!</div>

JohnRoesler:
### What does this do?
Jobs() needs to return the run now channel in order to enable jobs to use RunNow

resolves #667 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -454,10 +454,11 @@ func (s *scheduler) now() time.Time {
 
 func (s *scheduler) jobFromInternalJob(in internalJob) job {
 	return job{
-		id:            in.id,
-		name:          in.name,
-		tags:          slices.Clone(in.tags),
-		jobOutRequest: s.jobOutRequestCh,
+		in.id,
+		in.name,
+		slices.Clone(in.tags),
+		s.jobOutRequestCh,
+		s.runJobRequestCh,
 	}
 }
 
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -1479,7 +1479,7 @@ func TestScheduler_RunJobNow(t *testing.T) {
 		{
 			"duration job - start immediately",
 			chDurationImmediate,
-			DurationJob(time.Second * 10),
+			DurationJob(time.Second * 5),
 			func() {
 				chDurationImmediate <- struct{}{}
 			},
@@ -1489,7 +1489,7 @@ func TestScheduler_RunJobNow(t *testing.T) {
 				),
 			},
 			func() time.Duration {
-				return 10 * time.Second
+				return 5 * time.Second
 			},
 			2,
 		},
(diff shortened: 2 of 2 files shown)


scheduler.go
```diff
@@ -454,10 +454,11 @@ func (s *scheduler) now() time.Time {
...
 	return job{
-		id:            in.id,
-		name:          in.name,
-		tags:          slices.Clone(in.tags),
-		jobOutRequest: s.jobOutRequestCh,
+		in.id,
+		in.name,
+		slices.Clone(in.tags),
+		s.jobOutRequestCh,
+		s.runJobRequestCh,
```
JohnRoesler, line 461: switched this to not used named struct parameters so a future build will break if a new field is added to job but not populated here

scheduler_test.go
```diff
@@ -1529,9 +1529,10 @@ func TestScheduler_RunJobNow(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			s := newTestScheduler(t)
 
-			j, err := s.NewJob(tt.j, NewTask(tt.fun), tt.opts...)
+			_, err := s.NewJob(tt.j, NewTask(tt.fun), tt.opts...)
 			require.NoError(t, err)
 
+			j := s.Jobs()[0]
```
JohnRoesler, line 1535: updating this test to validate the exact behavior seen in the issue.

Issue #695, opened 2024-03-18
[BUG] -  RemoveJob almost always returns an error: gocron: job not found 


rigens:
### Describe the bug

Attempt to remove a scheduled jobs by ID fails

### To Reproduce

```go
func main() {
	scheduler, err := gocron.NewScheduler(gocron.WithLocation(time.UTC))
	if err != nil {
		log.Fatal(err)
	}

	var jobIds []uuid.UUID
	offset := time.Now().UTC().Add(32 * time.Second)

	for i := 0; i < 1000; i++ {
		runAt := offset.Add(time.Duration(i) * time.Second)
		job, err := scheduler.NewJob(
			gocron.OneTimeJob(gocron.OneTimeJobStartDateTime(runAt)),
			gocron.NewTask(func() {
				log.Printf("running job at %s", time.Now().UTC())
			}),
		)
		if err != nil {
			log.Fatal(err)
		}
		jobIds = append(jobIds, job.ID())
	}

	scheduler.Start()

	time.Sleep(2 * time.Second)

	for _, jobId := range jobIds {
		err := scheduler.RemoveJob(jobId)
		if err != nil {
			log.Printf("remove job %s failed: %s", jobId, err)
		} else {
			log.Printf("job %s removed", jobId)
		}
	}

	for {
		time.Sleep(time.Hour)
	}
}
```

### Version

v2.2.6

### Expected behavior

All scheduled jobs should be removed. But we have the output 

```
...
2024/03/18 11:41:00 remove job d518cb7c-2338-4481-8c4a-cd87a3a385d1 failed: gocron: job not found
2024/03/18 11:41:00 remove job a9b0e5ea-8ef7-4ce9-b738-2851da5280e3 failed: gocron: job not found
2024/03/18 11:41:00 remove job d75e5f3a-791f-4fa9-a515-6cfc12e90002 failed: gocron: job not found
2024/03/18 11:41:00 remove job f4ecdb50-ef6d-49b7-86c4-c0b053e6ed1e failed: gocron: job not found
2024/03/18 11:41:00 remove job d2d6daae-a914-41c4-81e5-240784f4f520 failed: gocron: job not found
2024/03/18 11:41:00 remove job c5960685-171e-4807-97d8-470901159ffb failed: gocron: job not found
2024/03/18 11:41:00 remove job f6424ac0-4c5d-4249-9ebb-8c2a0dc952eb failed: gocron: job not found
[...]

PR #699, opened 2024-03-22
Allow more time for requestJobCtx

drewgonzales360:
### What does this do?

If you attempted to get a job and the channel was full, you were returned a `nil`. Instead, this code will block and wait until a job is returned or the context deadline is exceeded.

https://github.com/go-co-op/gocron/issues/695

Yeah.

- [ ] Updated `README.md`

### DIFF
--- a/util.go
+++ b/util.go
@@ -45,17 +45,11 @@ func requestJob(id uuid.UUID, ch chan jobOutRequest) *internalJob {
 func requestJobCtx(ctx context.Context, id uuid.UUID, ch chan jobOutRequest) *internalJob {
 	resp := make(chan internalJob, 1)
 	select {
-	case <-ctx.Done():
-		return nil
-	default:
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -44,18 +44,12 @@ func requestJob(id uuid.UUID, ch chan jobOutRequest) *internalJob {
 
 func requestJobCtx(ctx context.Context, id uuid.UUID, ch chan jobOutRequest) *internalJob {
```
drewgonzales360, line 45: IMO, this function should throw an error if the context deadline is exceeded instead of returning `nil`.

JohnRoesler (approved):

util.go
JohnRoesler: yeah that seems reasonable. A context deadline exceeded error?

PR #705, opened 2024-03-29
fix nextRun with singleton mode reporting incorrect time

### LINKED ISSUE #704: [BUG] - Time reported by `NextRun` when running `WithSingletonMode` and `WithStartDateTime` incorrect
### Describe the bug
The time reported by `NextRun` isn't right when running `WithSingletonMode` and `WithStartDateTime`. For instance, suppose we have a job that starts at time `t` (in the future) and repeats with some interval `i`. When calling `NextRun` at some time `t_i` that is before `t`, the `NextRun` is reported correctly. However, after the job is run for the first time, the time reported by `NextRun` is `t + 2i` when it should be `t + i` instead. Here's a table depicting the pattern I observed

| Interval when `NextRun` is called      | Actual Return Value | Expected Return Value |
| ----------- | ----------- | ----------- |
| `0 -> t`      | `t`  ✅       | `t`   |
| `t -> t + i`   | `t + 2i` ❌         | `t + i`   |
| `t + i -> t + 2i`   | `t + 4i` ❌        | `t + 2i`   |

### To Reproduce

Here's a test that I wrote in the `job_test.go` file:

```go
func TestJob_NextRun(t *testing.T) {
	testTime := time.Now()

	s := newTestScheduler(t)

        // run a job every 10 milliseconds that starts 10 milliseconds after the current time
	j, err := s.NewJob(
		DurationJob(
			10*time.Millisecond,
		),
		NewTask(
			func() {},
		),
		WithStartAt(WithStartDateTime(testTime.Add(10*time.Millisecond))),
		WithSingletonMode(LimitModeReschedule),
	)
	require.NoError(t, err)

	s.Start()
	nextRun, err := j.NextRun()
	require.NoError(t, err)

        // `NextRun` should report `testTime.Add(10*time.Millisecond)`
	assert.Equal(t, testTime.Add(10*time.Millisecond), nextRun)

        // sleep for 11ms to to wait for the next job
	time.Sleep(11 * time.Millisecond)

	nextRun, err = j.NextRun()
	assert.NoError(t, err)

        // `NextRun` should report a time 20 milliseconds after `testTime`, but instead reports a value that is `30ms` after
	assert.Equal(t, testTime.Add(20*time.Millisecond), nextRun)
	assert.Equal(t, 20*time.Millisecond, nextRun.Sub(testTime))

	err = s.Shutdown()
	require.NoError(t, err)
}
```

### Version
`v2.2.9`

### Expected behavior
I expect that these two asserts:

```go
	assert.Equal(t, testTime.Add(20*time.Millisecond), nextRun)
	assert.Equal(t, 20*time.Millisecond, nextRun.Sub(testTime))
```

should not fail. I also modified this test to see what happens when more jobs are run:

```go
func TestJob_NextRun(t *testing.T) {
	testTime := time.Now()

	s := newTestScheduler(t)

        // run a job every 10 milliseconds that starts 10 milliseconds after the current time
	j, err := s.NewJob(
		DurationJob(
			10*time.Millisecond,
		),
		NewTask(
			func() {},
		),
		WithStartAt(WithStartDateTime(testTime.Add(10*time.Millisecond))),
		WithSingletonMode(LimitModeReschedule),
	)
	require.NoError(t, err)

	s.Start()
	nextRun, err := j.NextRun()
	require.NoError(t, err)

        // `NextRun` should report `testTime.Add(10*time.Millisecond)`
	assert.Equal(t, testTime.Add(10*time.Millisecond), nextRun)

        // sleep for 11ms to to wait for the next job
	time.Sleep(11 * time.Millisecond)

	nextRun, err = j.NextRun()
	assert.NoError(t, err)

        // `NextRun` should report a time 20 milliseconds after `testTime`, but instead reports a value that is `30ms` after
	assert.Equal(t, testTime.Add(20*time.Millisecond), nextRun)
	assert.Equal(t, 20*time.Millisecond, nextRun.Sub(testTime))

        // sleep for 11ms to to wait for the next job
	time.Sleep(11 * time.Millisecond)

	nextRun, err = j.NextRun()
	assert.NoError(t, err)

         // `NextRun` should report a time 30 milliseconds after `testTime`, but instead reports a value that is `50ms` after
	assert.Equal(t, testTime.Add(30*time.Millisecond), nextRun)
	assert.Equal(t, 30*time.Millisecond, nextRun.Sub(testTime))

	err = s.Shutdown()
	require.NoError(t, err)
```

Fortunately what is working: my job is running at the specified interval provided to `DurationJob`.

JohnRoesler:
### What does this do?
adjusts how the nextScheduledRun field is tracked. Rather than being a single time.Time field it is now a slice of time.Time. This allows for the case, that will often happen with limited/singleton jobs that may be queued up. In this case, there are multiple nextScheduledRun values.

resolves #704 

yes - thank you @aneesh1 for the excellent ticket write up and test case

- [ ] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -24,8 +24,12 @@ type internalJob struct {
 	name   string
 	tags   []string
 	jobSchedule
-	lastScheduledRun   time.Time
-	nextScheduled      time.Time
+	lastScheduledRun time.Time
+
+	// as some jobs may queue up, it's possible to
+	// have multiple nextScheduled times
+	nextScheduled []time.Time
+
 	lastRun            time.Time
 	function           any
 	parameters         []any
@@ -894,7 +898,12 @@ func (j job) NextRun() (time.Time, error) {
 	if ij == nil || ij.id == uuid.Nil {
 		return time.Time{}, ErrJobNotFound
 	}
-	return ij.nextScheduled, nil
+	if len(ij.nextScheduled) == 0 {
+		return time.Time{}, nil
+	}
+	// the first element is the next scheduled run with subsequent
+	// runs following after in the slice
+	return ij.nextScheduled[0], nil
 }
 
 func (j job) Tags() []string {
--- a/job.go
+++ b/job.go
@@ -24,7 +24,6 @@ type internalJob struct {
 	name   string
 	tags   []string
 	jobSchedule
-	lastScheduledRun time.Time
 
 	// as some jobs may queue up, it's possible to
 	// have multiple nextScheduled times
--- a/job_test.go
+++ b/job_test.go
@@ -492,3 +492,42 @@ func TestWithEventListeners(t *testing.T) {
 		})
 	}
 }
+
+func TestJob_NextRun(t *testing.T) {
+	testTime := time.Now()
+
+	s := newTestScheduler(t)
+
+	// run a job every 10 milliseconds that starts 10 milliseconds after the current time
+	j, err := s.NewJob(
+		DurationJob(
+			10*time.Millisecond,
+		),
+		NewTask(
+			func() {},
+		),
+		WithStartAt(WithStartDateTime(testTime.Add(10*time.Millisecond))),
+		WithSingletonMode(LimitModeReschedule),
+	)
+	require.NoError(t, err)
+
+	s.Start()
+	nextRun, err := j.NextRun()
+	require.NoError(t, err)
+
+	// `NextRun` should report `testTime.Add(10*time.Millisecond)`
+	assert.Equal(t, testTime.Add(10*time.Millisecond), nextRun)
+
+	// sleep for 11ms to wait for the next job
+	time.Sleep(11 * time.Millisecond)
+
+	nextRun, err = j.NextRun()
+	assert.NoError(t, err)
+
+	// `NextRun` should report a time 20 milliseconds after `testTime`, but instead reports a value that is `30ms` after
+	assert.Equal(t, testTime.Add(20*time.Millisecond), nextRun)
+	assert.Equal(t, 20*time.Millisecond, nextRun.Sub(testTime))
+
+	err = s.Shutdown()
+	require.NoError(t, err)
+}
--- a/job_test.go
+++ b/job_test.go
@@ -494,37 +494,57 @@ func TestWithEventListeners(t *testing.T) {
 }
 
 func TestJob_NextRun(t *testing.T) {
-	testTime := time.Now()
-
-	s := newTestScheduler(t)
-
-	// run a job every 10 milliseconds that starts 10 milliseconds after the current time
-	j, err := s.NewJob(
-		DurationJob(
-			100*time.Millisecond,
-		),
-		NewTask(
+	tests := []struct {
+		name string
+		f    func()
+	}{
+		{
+			"simple",
 			func() {},
-		),
-		WithStartAt(WithStartDateTime(testTime.Add(100*time.Millisecond))),
-		WithSingletonMode(LimitModeReschedule),
-	)
-	require.NoError(t, err)
+		},
+		{
+			"sleep 3 seconds",
+			func() {
+				time.Sleep(300 * time.Millisecond)
+			},
+		},
+	}
 
-	s.Start()
-	nextRun, err := j.NextRun()
-	require.NoError(t, err)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			testTime := time.Now()
 
-	assert.Equal(t, testTime.Add(100*time.Millisecond), nextRun)
+			s := newTestScheduler(t)
 
-	time.Sleep(150 * time.Millisecond)
+			// run a job every 10 milliseconds that starts 10 milliseconds after the current time
+			j, err := s.NewJob(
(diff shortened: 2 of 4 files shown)


scheduler.go
```diff
@@ -298,7 +298,12 @@ func (s *scheduler) selectExecJobsOutForRescheduling(id uuid.UUID) {
 		// so we don't need to reschedule it.
 		return
 	}
-	j.lastScheduledRun = j.nextScheduled
+	if len(j.nextScheduled) > 0 {
+		// always grab the last element in the slice as that is the furthest
+		// out in the future and the time from which we want to calculate
+		// the subsequent next run time.
+		j.lastScheduledRun = j.nextScheduled[len(j.nextScheduled)-1]
+	}
```
aneesh1: There's an assumption here that the jobs in `j.nextScheduled` are sorted. I understand that the code may be written in such a way that sorting is not needed, but it _might_ be a good idea to sort `nextScheduled` and then get the last element in the slice. What do you think?

JohnRoesler: Definitely a safe thing to do 👍

scheduler_test.go
```diff
@@ -1622,7 +1622,7 @@ func TestScheduler_RunJobNow(t *testing.T) {
 				WithSingletonMode(LimitModeReschedule),
 			},
 			func() time.Duration {
-				return 20 * time.Second
+				return 10 * time.Second
```
JohnRoesler, line 1625: this is correct now!

aneesh1:
(If you already have a test for this, feel free to ignore) Do we have / can we add a test where a job runs longer than the specified duration and the next job has to wait to highlight the expected behavior?

Ex: 
- Add DurationJob(5 * time.Second)
- t0: startJob (expecting job to run again at t5)
- t5: job is still running, so it’s skipped
- t7: job finishes
- t10: next job starts

cc @drewgonzales360

job.go
```diff
@@ -24,8 +24,12 @@ type internalJob struct {
 	name   string
 	tags   []string
 	jobSchedule
-	lastScheduledRun   time.Time
-	nextScheduled      time.Time
+	lastScheduledRun time.Time
+
+	// as some jobs may queue up, it's possible to
+	// have multiple nextScheduled times
+	nextScheduled []time.Time
```
drewgonzales360, line 30: If a job is running over and singleton mode is enabled, the job at the front of this list will get skipped. Do you plan to pop it off the front like a queue? If so, using a queue would be a nice abstraction here.

scheduler.go
```diff
@@ -298,7 +298,15 @@ func (s *scheduler) selectExecJobsOutForRescheduling(id uuid.UUID) {
...
 	}
-	j.lastScheduledRun = j.nextScheduled
+	if len(j.nextScheduled) > 0 {
+		// always grab the last element in the slice as that is the furthest
+		// out in the future and the time from which we want to calculate
+		// the subsequent next run time.
+		slices.SortStableFunc(j.nextScheduled, func(a, b time.Time) int {
+			return a.Compare(b)
+		})
+		j.lastScheduledRun = j.nextScheduled[len(j.nextScheduled)-1]
```
drewgonzales360: Is this being sorted by increasing time? Do you want the first element in this list?

scheduler.go
```diff
@@ -340,6 +348,19 @@ func (s *scheduler) selectExecJobsOutCompleted(id uuid.UUID) {
 		return
 	}
 
+	// if the job has more than one nextScheduled time,
+	// we need to remove any that are in the past.
+	if len(j.nextScheduled) > 1 {
```
drewgonzales360, line 354: Wouldn't you want to do this all the time? Even if `nextScheduled` had one element, and the timestamp on the one element was in the past, wouldn't you skip that execution?

job_test.go
```diff
@@ -492,3 +492,39 @@ func TestWithEventListeners(t *testing.T) {
...
+
+	s := newTestScheduler(t)
+
+	// run a job every 10 milliseconds that starts 10 milliseconds after the current time
+	j, err := s.NewJob(
+		DurationJob(
+			100*time.Millisecond,
+		),
+		NewTask(
+			func() {},
```
drewgonzales360:
```suggestion
			func() { time.Sleep(3 * time.Second) },
```

Could you also add some tests where this sleeps through _several_ of the next jobs that would run.

drewgonzales360 (commented):
Thanks for all the hard work on this 🙏

scheduler.go
JohnRoesler: So, this doesn't impact scheduling here because it's always based on the the diff from the last nextScheduled value. I had a reason to keep it if there was only 1 value, which was so that calling Job.NextRun would always return something, but a next run in the past is pretty useless.

scheduler.go
JohnRoesler: it's sorted increasing, so `now, now+1hr, now+2hr`. We're scheduling the next run here, so we want the greatest time in the slice, the last value.

JohnRoesler:
When you all could, please validate it's working as expected for you: [v2.2.10-rc1](https://github.com/go-co-op/gocron/releases/tag/v2.2.10-rc1)

Issue #710, opened 2024-04-19
[FEATURE] - Add Distributed Locker At Job level


0x01F4:
I don't see a way to have locker at job level it is only at Scheduler.

If Locker is only at scheduler level there is no way to pass/init locker with job level details.

For example: If we want to have different lock name for each it is not possible.

PR #711, opened 2024-04-19
Added Distributed Locker to JobOptions

0x01F4:
### What does this do?

It adds Distributed Locker At Job level.So that we can have different locker for each job

Support #710

No

Yes

- [x] Updated `README.md`


JohnRoesler (approved):
Very nice

PR #737, opened 2024-06-20
issue-736: moving validation of one-time to

### LINKED ISSUE #736: [BUG] - OneTimeJobStartDateTime returns error if date in the past
### Describe the bug

One of the nice features about the lib is that it has good support for testing using `github.com/jonboulle/clockwork` for simulating time. 

The method above instead hardcodes `time.Now()` and makes it harder for testing as even if we define a fake `clockwork.NewFakeClockAt` during the creation of the scheduler, the method above still hardcodes time.Now and leads to breaking tests.

### Version

v2.5.0

### Expected behavior

Either:

1. bypass the check
2. include a clockwork argument:

```go
// OneTimeJobStartDateTimeWithClock sets the date & time at which the job should run.
// This datetime must be in the future.
func OneTimeJobStartDateTimeWithClock(start time.Time, clock clockwork.Clock) OneTimeJobStartAtOption {
	return func(j *internalJob) error {
		if start.IsZero() || start.Before(clock.Now()) {
			return ErrOneTimeJobStartDateTimePast
		}
		j.startTime = start
		return nil
	}
}
```

rbroggi:

Resolves #736 

Yes

- [ ] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -98,7 +98,7 @@ type limitRunsTo struct {
 // JobDefinition defines the interface that must be
 // implemented to create a job from the definition.
 type JobDefinition interface {
-	setup(*internalJob, *time.Location) error
+	setup(j *internalJob, l *time.Location, now time.Time) error
 }
 
 var _ JobDefinition = (*cronJobDefinition)(nil)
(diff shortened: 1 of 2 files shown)


rbroggi:
the failed tests seem to be deriving from test-flakiness that is not linked to this change...

JohnRoesler (approved):
This is exactly what I was thinking 👍

PR #741, opened 2024-06-21
issue-740: expand oneTimeJob to support multiple times

### LINKED ISSUE #740: [FEATURE] - Support for an `AtTimes` job

In some instances I would like to schedule future jobs to be run at specific times. Those could be totally or only partially in the future (in case of an engine restart in the middle of the schedule). For convenience reasons, it can be way easier to define an `AtTimes` job rather than a dedicated cron-schedule if the input from another system is a limited list of times.

a new type of job that can be invoked similarly to the other, something like:

```go
// OneTimeJob is to run a job once at a specified time and not on
// any regular schedule.
func AtTimesTimeJob(atTimes ...time.Time) JobDefinition
```

Multiple schedules with `OneTimeJob` is an alternative to the above. I find it less convenient and worse case the `AtTimes` should be a more generic way to implement also the `OneTimeJob`.

rbroggi:

Resolves #740 

- [ ] Updated `README.md`

### DIFF
--- a/job.go
+++ b/job.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"sort"
 	"strings"
 	"time"
 
@@ -486,6 +487,46 @@ func OneTimeJob(startAt OneTimeJobStartAtOption) JobDefinition {
 	}
 }
 
+var _JobDefinition = (*atTimesJobDefinition)(nil)
+
+type atTimesJobDefinition struct {
+	atTimes []time.Time
+}
+
+func (a atTimesJobDefinition) setup(j *internalJob, _ *time.Location, now time.Time) error {
+	sortedTimes := a.atTimes
+	sort.Slice(a.atTimes, func(i, j int) bool {
+		return a.atTimes[i].Before(a.atTimes[j])
+	})
+	// keep only schedules that are in the future
+	idx, found := slices.BinarySearchFunc(sortedTimes, now, timeCmp())
+	if found {
+		idx++
+	}
+	sortedTimes = sortedTimes[idx:]
+	if len(sortedTimes) == 0 {
+		return ErrAtTimesJobAtLeastOneInFuture
+	}
+	j.jobSchedule = atTimesJob{sortedTimes: sortedTimes}
+	return nil
+}
+
+func timeCmp() func(element time.Time, target time.Time) int {
+	return func(element time.Time, target time.Time) int {
+		if element.Equal(target) {
+			return 0
+		}
+		if element.Before(target) {
+			return -1
+		}
+		return 1
+	}
+}
+
+func AtTimesJob(atTimes ...time.Time) JobDefinition {
+	return atTimesJobDefinition{atTimes: atTimes}
+}
+
 // -----------------------------------------------
 // -----------------------------------------------
 // ----------------- Job Options -----------------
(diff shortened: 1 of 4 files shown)


rbroggi:
While implementing this functionality I found unexpected that the `NextRun` will not return `time.Time{}` upon exhausted runs. 

I debugged the implementation and saw that the `next` method implemented by the new job works as expected - returning `time.Time{}` upon input `lastRun` that are after the sortedTimes. 

This could be a potential bug also for the cronJob schedule - I wonder what it returns upon schedule exhaustion.

I have issued a PR to demonstrate this concern for the OneTimeJob: #742

JohnRoesler:
hey @rbroggi thanks for all your contributions lately!

For the implementation on this one, I think it would be nice and possible to utilize the existing OneTimeJob but expand it to allow multiple runs.

Something like (incomplete code...)

```golang
var _ JobDefinition = (*oneTimeJobDefinition)(nil)

type oneTimeJobDefinition struct {
	startAt OneTimeJobStartAtOption
}

func (o oneTimeJobDefinition) setup(j *internalJob, _ *time.Location, now time.Time) error {
	var oneTimeJobSchedule oneTimeJob
	atTimes, err := o.startAt(j)
	if err != nil {
		return err
	}
	parsedAtTimes := make([]time.Time, 0)
	for _, at := range atTimes {
		if at.After(now) {
			parsedAtTimes = append(parsedAtTimes, at)
		}
	}
	slices.SortStableFunc(parsedAtTimes, func(a, b time.Time) int {
		return a.Compare(b)
	})
	if len(parsedAtTimes) != 0 {
		j.startTime = parsedAtTimes[0]
		oneTimeJobSchedule.atTimes = parsedAtTimes
	}
	j.jobSchedule = oneTimeJobSchedule
	
	// in case we are not in the `startImmediately` case, our start-date must be in
	// the future according to the scheduler clock
	if !j.startImmediately && (j.startTime.IsZero() || j.startTime.Before(now)) {
		return ErrOneTimeJobStartDateTimePast
	}
	return nil
}

// OneTimeJobStartAtOption defines when the one time job is run
type OneTimeJobStartAtOption func(*internalJob) ([]time.Time, error)

// OneTimeJobStartImmediately tells the scheduler to run the one time job immediately.
func OneTimeJobStartImmediately() OneTimeJobStartAtOption {
	return func(j *internalJob) ([]time.Time, error) {
		j.startImmediately = true
		return nil, nil
	}
}

// OneTimeJobStartDateTime sets the date & time at which the job should run.
// This datetime must be in the future (according to the scheduler clock).
func OneTimeJobStartDateTime(start time.Time) OneTimeJobStartAtOption {
	return func(j *internalJob) ([]time.Time, error) {
		j.startTime = start
		return nil, nil
	}
}

// OneTimeJobStartDateTimes sets the date & times at which the job should run.
// At least one of the date/times must be in the future (according to the scheduler clock).
func OneTimeJobStartDateTimes(times []time.Time) OneTimeJobStartAtOption {
	return func(j *internalJob) ([]time.Time, error) {
		if len(times) == 0 {
			return nil, fmt.Errorf("TODO")
		}
		return times, nil
	}
}

// OneTimeJob is to run a job once at a specified time and not on
// any regular schedule.
func OneTimeJob(startAt OneTimeJobStartAtOption) JobDefinition {
	return oneTimeJobDefinition{
		startAt: startAt,
	}
}





var _ jobSchedule = (*oneTimeJob)(nil)

type oneTimeJob struct {
	atTimes []time.Time
}

func (o oneTimeJob) next(lastRun time.Time) time.Time {
	if len(o.atTimes) != 0 {
		next := o.atTimes[0]
		o.atTimes = o.atTimes[1:]
		if !next.After(lastRun) {
			return time.Time{}
		}
		return next
	}
	return time.Time{}
}
```

What do you think about that? I haven't tested or anything, but I think it should work.

rbroggi:
I think it should work as well, it sounds a bit misleading to me to have a `OneTimeJob` actually implementing a `ManyTimesJob` 
but that's only my opinion, happy to accommodate your suggestion if you think it aligns better with the lib design.

JohnRoesler:
@rbroggi I do agree, the name is a little misleading, but I think updating the description of the OneTimeJob will suffice. It's really more like "limited set of job(s)". I like how it fits vs. adding an entirely new job type. 🤝

rbroggi:
> @rbroggi I do agree, the name is a little misleading, but I think updating the description of the OneTimeJob will suffice. It's really more like "limited set of job(s)". I like how it fits vs. adding an entirely new job type. 🤝

Awesome, will followup! thank you for the fast feedback :heart:

job.go
```diff
@@ -446,35 +447,47 @@ type oneTimeJobDefinition struct {
...
+	if !j.startImmediately && len(sortedTimes) == 0 {
 		return ErrOneTimeJobStartDateTimePast
 	}
+	j.jobSchedule = oneTimeJob{sortedTimes: sortedTimes}
 	return nil
 }
 
 // OneTimeJobStartAtOption defines when the one time job is run
-type OneTimeJobStartAtOption func(*internalJob) error
+type OneTimeJobStartAtOption func(*internalJob) []time.Time
```
rbroggi, line 468: I had to change the signature of this exported type - I tried to think about situations where this would not be considered a backwards-compatible change but I could not think of one.

job.go
```diff
@@ -446,35 +447,47 @@ type oneTimeJobDefinition struct {
 }
 
 func (o oneTimeJobDefinition) setup(j *internalJob, _ *time.Location, now time.Time) error {
-	j.jobSchedule = oneTimeJob{}
-	if err := o.startAt(j); err != nil {
-		return err
+	sortedTimes := o.startAt(j)
+	sort.Slice(sortedTimes, func(i, j int) bool {
+		return sortedTimes[i].Before(sortedTimes[j])
+	})
```
JohnRoesler, line 453: I've used this in other places, Before vs. Compare. SortStable, which isn't super important here, just keeps the order if items are duplicates
```golang
slices.SortStableFunc(j.nextScheduled, func(a, b time.Time) int {
			return a.Compare(b)
		})
```

JohnRoesler (approved):
This looks good 👍 I tried to think of any edge cases with input, but you handle if they pass an empty slice, or times in the past

PR #744, opened 2024-06-22
issue-740: ascending time function

rbroggi:

Followup optional comments on #740 and including test for out of order job `atTimes` which were missing.

Yes

- [x] Updated `README.md`

### DIFF
--- a/util.go
+++ b/util.go
@@ -88,11 +88,11 @@ func convertAtTimesToDateTime(atTimes AtTimes, location *time.Location) ([]time.
 		}
 		atTimesDate = append(atTimesDate, at.time(location))
 	}
-	slices.SortStableFunc(atTimesDate, AscendingTime)
+	slices.SortStableFunc(atTimesDate, ascendingTime)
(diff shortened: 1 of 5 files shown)


util.go
```diff
@@ -88,12 +88,14 @@ func convertAtTimesToDateTime(atTimes AtTimes, location *time.Location) ([]time.
...
 		atTimesDate = append(atTimesDate, at.time(location))
 	}
-	slices.SortStableFunc(atTimesDate, func(a, b time.Time) int {
-		return a.Compare(b)
-	})
+	slices.SortStableFunc(atTimesDate, AscendingTime)
 	return atTimesDate, nil
 }
 
+func AscendingTime(a, b time.Time) int {
```
JohnRoesler: This shouldn't need to be exported

rbroggi: good catch, I didn't realize that it was all in the same package. Pushed :+1:

JohnRoesler (approved):

PR #752, opened 2024-07-09
issue-751: Address Bug in Rescheduling

### LINKED ISSUE #751: [BUG] - Reschedule Count Up From Time Zero
### Describe the bug

A state can be reached such that `scheduleFrom` in `selectExecJobsOutForRescheduling` retains the zero value, and this loop executes counting up from time zero to the current time:

```
	if next.Before(s.now()) {
		// in some cases the next run time can be in the past, for example:
		// - the time on the machine was incorrect and has been synced with ntp
		// - the machine went to sleep, and woke up some time later
		// in those cases, we want to increment to the next run in the future
		// and schedule the job for that time.
		for next.Before(s.now()) {
			next = j.next(next)
		}
	}
```
https://github.com/go-co-op/gocron/blob/970c5399bc9de7efb49e51207aa034ea503d7b4f/scheduler.go#L309-L333

### To Reproduce

Cause a job with a set run interval  to be sent out for rescheduling when `j.nextScheduled` is an empty slice

### Version

v2.8.0

### Expected behavior

The computation for next iteration starts at a more reasonable time such as `j.lastRun` or `j.startTime` if defined, when `j.nextScheduled` is empty

samuelattwood:

Resolves #751 

- [ ] Updated `README.md`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -306,7 +306,8 @@ func (s *scheduler) selectExecJobsOutForRescheduling(id uuid.UUID) {
 		// so we don't need to reschedule it.
 		return
 	}
-	var scheduleFrom time.Time
+
+	scheduleFrom := j.lastRun
 	if len(j.nextScheduled) > 0 {
 		// always grab the last element in the slice as that is the furthest
 		// out in the future and the time from which we want to calculate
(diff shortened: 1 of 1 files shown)


JohnRoesler (approved):
👍 thanks for the improvement

PR #754, opened 2024-07-10
feat: add WithIdentifier

### LINKED ISSUE #753: [FEATURE] - Allow UUID to be set in the job
I have https://github.com/LumeWeb/gocron/commit/6e1fcc64e8b6a5b4ae0025ac71f25a3e7b67a79b though im not sure if you require anything else.

would like to submit a PR so I don't need to keep a fork. Its a small detail but important for my use case.

Thanks.

@JohnRoesler

pcfreak30:
### What does this do?

Adds an job definition builder option to set the UUID

Resolves #753

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -38,6 +38,7 @@ var (
 	ErrWithDistributedElectorNil     = fmt.Errorf("gocron: WithDistributedElector: elector must not be nil")
 	ErrWithDistributedLockerNil      = fmt.Errorf("gocron: WithDistributedLocker: locker must not be nil")
 	ErrWithDistributedJobLockerNil   = fmt.Errorf("gocron: WithDistributedJobLocker: locker must not be nil")
+	ErrWithIdentifierNil             = fmt.Errorf("gocron: WithIdentifier: identifier must not be nil")
 	ErrWithLimitConcurrentJobsZero   = fmt.Errorf("gocron: WithLimitConcurrentJobs: limit must be greater than 0")
 	ErrWithLocationNil               = fmt.Errorf("gocron: WithLocation: location must not be nil")
 	ErrWithLoggerNil                 = fmt.Errorf("gocron: WithLogger: logger must not be nil")
--- a/example_test.go
+++ b/example_test.go
@@ -694,6 +694,27 @@ func ExampleWithGlobalJobOptions() {
 	// [tag4 tag5 tag6]
 }
 
+func ExampleWithIdentifier() {
+	s, _ := NewScheduler()
+	defer func() { _ = s.Shutdown() }()
+
+	j, _ := s.NewJob(
+		DurationJob(
+			time.Second,
+		),
+		NewTask(
+			func(one string, two int) {
+				fmt.Printf("%s, %d", one, two)
+			},
+			"one", 2,
+		),
+		WithIdentifier(uuid.MustParse("87b95dfc-3e71-11ef-9454-0242ac120002")),
+	)
+	fmt.Println(j.ID())
+	// Output:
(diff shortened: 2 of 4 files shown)


JohnRoesler (dismissed):

job.go
```diff
@@ -610,6 +610,16 @@ func WithTags(tags ...string) JobOption {
 	}
 }
 
+// WithIdentifier sets the identifier for the job. The identifier
+// is used to uniquely identify the job and is used for logging
+// and metrics.
+func WithIdentifier(id uuid.UUID) JobOption {
+	return func(j *internalJob) error {
+		j.id = id
```
JohnRoesler, line 622: this should check if the id passed in is nil and return an error if it is. also a test would be a grand thing to have. 
- errors go here -> https://github.com/go-co-op/gocron/blob/399ac2807bb8cfb0cade9a73f6a8d48d9ba45f19/errors.go#L41
- error occurred tests here -> https://github.com/go-co-op/gocron/blob/64d6e48dae23488d9a160ae3b1f5070e3c06228a/scheduler_test.go#L502

pcfreak30: @JohnRoesler ok... however uuid.UUID is a struct and it isn't passed as a pointer/ref, so you can't test it for nil.

JohnRoesler: I meant uuid.Nil, which tests for an uninitialized uuid.

pcfreak30: @JohnRoesler check now.

JohnRoesler: thank you

example_test.go
```diff
@@ -694,6 +694,27 @@ func ExampleWithGlobalJobOptions() {
...
+			func(one string, two int) {
+				fmt.Printf("%s, %d", one, two)
+			},
+			"one", 2,
+		),
+		WithIdentifier(uuid.MustParse("87b95dfc-3e71-11ef-9454-0242ac120002")),
+	)
+	fmt.Println(j.ID())
+	// Output:
+	// [tag1 tag2 tag3]
```
JohnRoesler: This is going to print the uuid, so I think it should be - just have to make sure that's how uuid's print 😄 
```suggestion
	// 87b95dfc-3e71-11ef-9454-0242ac120002
```

errors.go
```diff
@@ -38,6 +38,7 @@ var (
 	ErrWithDistributedElectorNil     = fmt.Errorf("gocron: WithDistributedElector: elector must not be nil")
 	ErrWithDistributedLockerNil      = fmt.Errorf("gocron: WithDistributedLocker: locker must not be nil")
 	ErrWithDistributedJobLockerNil   = fmt.Errorf("gocron: WithDistributedJobLocker: locker must not be nil")
+	ErrWithWithIdentifierNil         = fmt.Errorf("gocron: WithIdentifier: identifier must not be nil")
```
JohnRoesler: Let's remove the extra `with` 
```suggestion
	ErrWithIdentifierNil         = fmt.Errorf("gocron: WithIdentifier: identifier must not be nil")
```

example_test.go
JohnRoesler: Yes should be that
```
--- FAIL: ExampleWithIdentifier (0.00s)
got:
87b95dfc-3e71-11ef-9454-0242ac120002
want:
[tag1 tag2 tag3]
```

pcfreak30: @JohnRoesler  fixed, i think.

errors.go
pcfreak30: woops :sweat_smile:

pcfreak30: @JohnRoesler fixed

JohnRoesler (approved):
cheers @pcfreak30

PR #755, opened 2024-07-11
issue-677: support task creation with variadic args

### LINKED ISSUE #677: [BUG] - Cannot create job with task with variadic args
### Describe the bug

When creating a job with a task that has variadic args, the validator in the library will expect the variadic args to be "1" argument (suspect to be a slice of that type). So creating a job with a task that has function of variadic arguments will always return `ErrNewJobWrongNumberOfParameters`. 

### To Reproduce

1. Create a function that has variadic args
2. Try create a cronjob from that

```go
scheduler, _ := gocron.NewScheduler()

myfunc := func(args ...string) {
  panic("This works")
}

// will have error
job, err := scheduler.NewJob(
  gocron.CronJob("* * * * 1", false),
  gocron.NewTask(myfunc), // the validator will expect num param to be 1 (slice of string)
)

```

### Version

v2.2.1

### Expected behavior

should be able to execute task with variadic parameters.

Higan:
### What does this do?
This pull request implements variadic args support for task creation

Resolves #677 

I have added several cases in `scheduler_test.go` based on my changes

- [x] Updated `scheduler_test.go`

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -535,6 +535,47 @@ func (s *scheduler) NewJob(jobDefinition JobDefinition, task Task, options ...Jo
 	return s.addOrUpdateJob(uuid.Nil, jobDefinition, task, options)
 }
 
+func (s *scheduler) verifyParameterType(taskFunc reflect.Value, tsk task) error {
+	isVariadic := taskFunc.Type().IsVariadic()
+
+	if isVariadic {
+		parameterType := taskFunc.Type().In(0).Elem().Kind()
+		if parameterType == reflect.Interface || parameterType == reflect.Pointer {
+			parameterType = reflect.Indirect(reflect.ValueOf(taskFunc.Type().In(0))).Kind()
+		}
+
+		for i := range tsk.parameters {
+			argumentType := reflect.TypeOf(tsk.parameters[i]).Kind()
+			if argumentType == reflect.Interface || argumentType == reflect.Pointer {
+				argumentType = reflect.TypeOf(tsk.parameters[i]).Elem().Kind()
+			}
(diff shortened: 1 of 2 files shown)


scheduler.go
```diff
@@ -535,6 +535,47 @@ func (s *scheduler) NewJob(jobDefinition JobDefinition, task Task, options ...Jo
 	return s.addOrUpdateJob(uuid.Nil, jobDefinition, task, options)
 }
 
+func (s *scheduler) verifyParameterType(taskFunc reflect.Value, tsk task) error {
+	isVariadic := taskFunc.Type().IsVariadic()
```
JohnRoesler: 🚀 epic, i was poking around the other day but didn't notice this IsVariadic method

scheduler_test.go
```diff
@@ -899,6 +899,26 @@ func TestScheduler_NewJobTask(t *testing.T) {
...
+			nil,
+		},
+		{
+			"all good struct - variadic",
+			NewTask(func(args ...interface{}) {}, struct{}{}),
+			nil,
+		},
+		{
+			"all good no arguments passed in - variadic",
+			NewTask(func(args ...interface{}) {}),
```
JohnRoesler, line 924: could you add a test case with multiple argument types where the last one is variadic? 

something like
```golang
NewTask(func(arg int, args ...string) {}, 1, "one", "two")
```

Higan: Sure this was neglected in my previous commit. And it's added now in this PR. Please have a look.

JohnRoesler (approved):

PR #757, opened 2024-07-17
issue-756: fix interface variadic parameters verifying

### LINKED ISSUE #756: [BUG] - cannot handle variadic interface parameters
This code does not work:

```go
func main() {
	j, err := s.NewJob(
		gocron.DurationJob(
			10*time.Second,
		),
		gocron.NewTask(
			func(parameters ...any) {
				for i := range parameters {
					fmt.Println(parameters[i])
				}
			},
			1,
			"2",
		),
	)
	if err != nil {
		panic(err)
	}
}
```
It panicked like this: 
image

This because the verifyVariadic only checks the first element's underlying type of `parameters ...any`, the first element's underlying type is `int` here and the second element's underlying type is `string`.

What I want: This code runs well without any panicking and keeps type-safe.

The correct behavior is that if the variadic parameter's type is interface, we only need to check if the actual parameters implement that interface.

I'd like to send a PR to fix this issue.

apocelipes:
Fixes #756.

### What does this do?
Check interface-implements if variadic parameter's types is an interface.

Resolves https://github.com/go-co-op/gocron/issues/756

Nothing.

Yes. I have added 2 test cases.

- [x] Updated `scheduler_test.go`

### DIFF
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -926,7 +926,7 @@ func TestScheduler_NewJobTask(t *testing.T) {
 			nil,
 		},
 		{
-			"all good no arguments passed in - interface variadic",
+			"all good - interface variadic, int, string",
 			NewTask(func(args ...interface{}) {}, 1, "2", 3.0),
 			nil,
 		},
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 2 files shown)


scheduler_test.go
```diff
@@ -924,6 +925,16 @@ func TestScheduler_NewJobTask(t *testing.T) {
 			NewTask(func(args ...interface{}) {}),
 			nil,
 		},
+		{
+			"all good no arguments passed in - interface variadic",
```
```suggestion
			"all good - interface variadic, int, string",
```

JohnRoesler (approved):

PR #760, opened 2024-07-18
issue-654: allow setting a stopTime for job.

### LINKED ISSUE #654: [FEATURE] Allow setting a stopTime on jobs to run the job until that date/time is reached
The v2 jobs requires an end time. Now it only has a start time. It would be better to add the end time to the attribute configuration.

Higan:
### What does this do?
This PR exposes a `stopTime` when creating a job so that the job cannot be executed once the `stopTime` is reached.

Resolves #654 

Yes I created several test cases in `scheduler_test.go` regarding job creation with `stopTime` specified and also the expected behavior when `stopTime` is reached.

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -146,6 +146,11 @@ func (e *executor) start() {
 						// safety check as it'd be strange bug if this occurred
 						return
 					}
+
+					if j.stopTimeReached() {
+						return
+					}
+
 					if j.singletonMode {
 						// for singleton mode, get the existing runner for the job
 						// or spin up a new one
--- a/executor.go
+++ b/executor.go
@@ -169,11 +169,6 @@ func (e *executor) start() {
 						// safety check as it'd be strange bug if this occurred
 						return
 					}
-
-					if j.stopTimeReached() {
-						return
-					}
-
 					if j.singletonMode {
 						// for singleton mode, get the existing runner for the job
 						// or spin up a new one
@@ -361,6 +356,10 @@ func (e *executor) runJob(j internalJob, jIn jobIn) {
 	case <-j.ctx.Done():
 		return
 	default:
+	}
+
+	if j.stopTimeReached(e.clock.Now()) {
+		return
 	}
 
 	if e.elector != nil {
--- a/job.go
+++ b/job.go
@@ -38,6 +38,7 @@ type internalJob struct {
 	limitRunsTo        *limitRunsTo
 	startTime          time.Time
 	startImmediately   bool
+	stopTime           time.Time
 	// event listeners
 	afterJobRuns          func(jobID uuid.UUID, jobName string)
 	beforeJobRuns         func(jobID uuid.UUID, jobName string)
@@ -58,6 +59,13 @@ func (j *internalJob) stop() {
 		j.timer.Stop()
 	}
 	j.cancel()
+}
+
+func (j *internalJob) stopTimeReached() bool {
+	if j.stopTime.IsZero() {
+		return false
+	}
+	return j.stopTime.Before(time.Now())
 }
 
 // task stores the function and parameters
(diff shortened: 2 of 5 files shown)


job.go
```diff
@@ -60,6 +61,13 @@ func (j *internalJob) stop() {
 	j.cancel()
 }
 
+func (j *internalJob) stopTimeReached() bool {
```
JohnRoesler: In order to support time mocking, this function should accept a now time.Time value. then, when it is called, we should pass in the schedulers clockwork time now value

executor.go
```diff
@@ -146,6 +146,11 @@ func (e *executor) start() {
 						// safety check as it'd be strange bug if this occurred
 						return
 					}
+
+					if j.stopTimeReached() {
+						return
+					}
```
JohnRoesler: In order to cover all of the job types, this check should happen inside the runJob() method of the executor right after it checks if the job or executor contexts are canceled

JohnRoesler: In order to support time mocking, we probably need to move the clockwork object off of the scheduler struct and into the executor struct, so that the clockwork time now value can be passed in here

job.go
```diff
@@ -595,11 +603,40 @@ func WithStartDateTime(start time.Time) StartAtOption {
...
+		if !j.stopTime.IsZero() && j.stopTime.Before(start) {
+			return ErrStartTimeLaterThanEndTime
+		}
 		j.startTime = start
 		return nil
 	}
 }
 
+// WithStopAt sets the option for stopping the job at
+// a specific datetime.
```
```suggestion
// WithStopAt sets the option for stopping the job from running
// after the specified time.
```

job.go
```diff
@@ -595,11 +603,40 @@ func WithStartDateTime(start time.Time) StartAtOption {
...
+	return func(j *internalJob) error {
+		return option(j)
+	}
+}
+
+// StopAtOption defines options for stopping the job
+type StopAtOption func(*internalJob) error
+
+// WithStopDateTime sets the final data & time after which the job should stop
+// This datetime must be in the future and should be after the startTime (if specified)
```
```suggestion
// WithStopDateTime sets the final date & time after which the job should stop.
// This must be in the future and should be after the startTime (if specified). 
// The job's final run may be at the stop time, but not after.
```

scheduler.go
```diff
@@ -473,6 +473,10 @@ func (s *scheduler) selectStart() {
 				next = j.next(s.now())
 			}
 
+			if j.stopTimeReached() {
```
JohnRoesler: This check shouldn't need to happen here, but instead in the `selectExecJobsOutForRescheduling` function. since the executor is checking, if the stop time is reached, new jobs will be caught by that and then not rescheduled.

JohnRoesler (commented):
Awesome work so far!

executor.go
JohnRoesler: I'll take care of moving the clock. I'm going to open a PR to do that.

job.go
```diff
@@ -595,11 +603,40 @@ func WithStartDateTime(start time.Time) StartAtOption {
...
+// WithStopAt sets the option for stopping the job at
+// a specific datetime.
+func WithStopAt(option StopAtOption) JobOption {
+	return func(j *internalJob) error {
+		return option(j)
+	}
+}
+
+// StopAtOption defines options for stopping the job
+type StopAtOption func(*internalJob) error
```
JohnRoesler: Once my PR is merged, could you pull it in and change the signature here to accept a time.Time as well, and pass that in to avoid using time.Now() for validation inside the WithStopDateTime func?

Higan:
@JohnRoesler  Sure I'll handle these later today. Thanks for reviewing and commenting.

Higan:
Hi @JohnRoesler I've made some changes with regard to your comments and now you could check this PR again. Thank you.

JohnRoesler (approved):

PR #776, opened 2024-09-16
fix: cleanup timers

hayotbisonai:
### What does this do?
This PR adds cleanup logic to handle existing timers in the `selectExecJobsOutForRescheduling` function of the scheduler. By stopping and clearing old timers, it prevents memory leaks and ensures that jobs are rescheduled correctly without accumulating unused timers.

- Added cleanup logic to stop and clear existing timers before setting a new timer.
- Ensured that old timers are properly disposed of to avoid memory leaks.

- [ ] Updated `README.md`

### Notes
- The changes focus on improving memory management by ensuring that previous timers are stopped and cleared.
- This update should help prevent potential memory leaks and ensure accurate job scheduling.


JohnRoesler (approved):
🧹 💯

PR #780, opened 2024-09-19
feat(monitor): introduce MonitorStatus

### LINKED ISSUE #775: [FEATURE] - Add JobStatus to RecordJobTiming in Monitor Interface
Morning fellas, really loving working with `gocron`, it allows me to quickly schedule repeating tasks! :) 

Im currently building a Prometheus Exporter and using the `Monitor` interface to measure task durations and add the results to a Prometheus Histogram. I would love to be able to also add whether a task was successful or not as a label to the metric.

I was wondering whether its possible to add the `JobStatus` to the `RecordJobTiming` function, similar as `IncrementJob` already has it in the `Monitor` interface.

I believe, we could pass the JobStatus, based on the error result when running `e.callJobWithRecover(j)` in:  
https://github.com/go-co-op/gocron/blob/c84d8f70c4038b86776433651a307c831e3daa78/executor.go#L399-L408

Im aware, that this is probably a breaking change, but just wanted to get your thoughts on that one.

FalcoSuessgott:
### What does this do?
In regard of #775, This PR introduces a new interface `MonitorStatus` that encapsulates `Monitor` but adds `RecordJobTimingStatus(startTime, endTime time.Time, id uuid.UUID, name string, tags []string, status JobStatus, err error)`, to allow recording a Job with its name, tags, start & end time, as well as the job status and error. I choose to add a new interface instead of changing `Monitor` to prevent introducing a breaking change.

Fixes #775 

none

- [ ] Updated `example_test.go` (`ExampleWithMonitor()` is commented out :D)
- [x] Updated `README.md`

### Notes
I believe having the `JobStatus` and `error` in `RecordJobTiming`, definitely make sense, in case you're fine with it we can also just extend `Monitor` even if it means having a breaking change ...


FalcoSuessgott:
Tested it on my end with `replace github.com/go-co-op/gocron/v2 v2.12.0 => github.com/FalcoSuessgott/gocron/v2 v2.12.2` in my `go.mod`. 

```go
// New returns a new scheduler.
func New(monitor gocron.MonitorStatus) Scheduler {
	s, err := gocron.NewScheduler(
		gocron.WithMonitorStatus(monitor),
		gocron.WithLogger(slog.Default()),
		gocron.WithGlobalJobOptions(
			gocron.WithSingletonMode(gocron.LimitModeReschedule),
		),
	)
	if err != nil {
		log.Fatalf("could not start scheduler: %v", err)
	}

	return Scheduler{s}
}
```

`MonitorStatus` implementation:

```go
func (c *Collector) RecordJobTimingWithStatus(startTime, endTime time.Time, uid uuid.UUID, name string, _ []string, status gocron.JobStatus, err error) {
	log.Printf("[%s (%s)] %s - %s => %v\n", name, uid, startTime, endTime, status)
}

func (c *Collector) RecordJobTiming(startTime, endTime time.Time, _ uuid.UUID, name string, _ []string) {}

func (c *Collector) IncrementJob(id uuid.UUID, name string, tags []string, status gocron.JobStatus) {}
```

And the output is exactly what I need:
```bash
....
2024/09/20 09:45:14 [vmware_vm (23dbda12-79ec-4f5d-989a-c240dd636639)] 2024-09-20 09:45:14.468317 +1000 AEST m=+13.162815626 - 2024-09-20 09:45:14.769485 +1000 AEST m=+13.463985126 => success
....

JohnRoesler (approved):

PR #813, opened 2025-01-08
feat(event listener): introduce BeforeJobRunsSkipIfBeforeFuncErrors as a new Eventlistener

FalcoSuessgott:
### What does this do?
This PR adds a new `EventListener` `BeforeJobRunsSkipIfBeforeFuncErrors` that exits job execution if the provided func returns an error

**Use Case**:
Im using `gocron` as a command wrapper for `terraform` to perform synthetic monitoring checks and came up with a configuration as follows:

```yaml
settings:
  interval: 10s
  extra_labels:
    development: true

tests:
  vmware_vm:
    env:
      TF_VAR_content: "This is a variable"

    before_script:
      - cmd: terraform init
        retries: 2

    script:
      - cmd: terraform apply -auto-approve

    after_script:
      - cmd: terraform destroy -auto-approve
```

Where as the commands specified in the `before_script` block would be passed to `BeforeJobRuns()`, without this PR there was no way to stop the currents jobs execution in case of an error in any of the commands in `before_script`. 

Introducing `BeforeJobRunsSkipIfBeforeFuncErrors` allows me to exit the jobs execution gracefully and reschedule to job after the given interval.

`none`

Si

- [ ] Updated `README.md`

### DIFF
--- a/.github/workflows/file_formatting.yml
+++ b/.github/workflows/file_formatting.yml
@@ -16,4 +16,4 @@ jobs:
         uses: actions/checkout@v4
       - name: verify example_test.go
         run: |
-          grep "^func " example_test.go | sort -c
+          grep "^func [a-z-A-Z]" example_test.go | sort -c
--- a/executor.go
+++ b/executor.go
@@ -389,7 +389,20 @@ func (e *executor) runJob(j internalJob, jIn jobIn) {
 		}
 		defer func() { _ = lock.Unlock(j.ctx) }()
 	}
+
 	_ = callJobFuncWithParams(j.beforeJobRuns, j.id, j.name)
+
+	err := callJobFuncWithParams(j.beforeJobRunsSkipIfBeforeFuncErrors, j.id, j.name)
+	if err != nil {
+		e.sendOutForRescheduling(&jIn)
+
+		select {
+		case e.jobsOutCompleted <- j.id:
+		case <-e.ctx.Done():
+		}
+
+		return
+	}
 
 	e.sendOutForRescheduling(&jIn)
 	select {
@@ -398,7 +411,6 @@ func (e *executor) runJob(j internalJob, jIn jobIn) {
 	}
 
 	startTime := time.Now()
-	var err error
 	if j.afterJobRunsWithPanic != nil {
 		err = e.callJobWithRecover(j)
 	} else {
--- a/job.go
+++ b/job.go
@@ -40,12 +40,13 @@ type internalJob struct {
 	startImmediately   bool
 	stopTime           time.Time
 	// event listeners
-	afterJobRuns          func(jobID uuid.UUID, jobName string)
-	beforeJobRuns         func(jobID uuid.UUID, jobName string)
-	afterJobRunsWithError func(jobID uuid.UUID, jobName string, err error)
-	afterJobRunsWithPanic func(jobID uuid.UUID, jobName string, recoverData any)
-	afterLockError        func(jobID uuid.UUID, jobName string, err error)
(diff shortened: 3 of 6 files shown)


executor.go
```diff
@@ -389,7 +389,27 @@ func (e *executor) runJob(j internalJob, jIn jobIn) {
...
 		defer func() { _ = lock.Unlock(j.ctx) }()
 	}
-	_ = callJobFuncWithParams(j.beforeJobRuns, j.id, j.name)
+
+	// execute before func, unless user has specified BeforeJobRunsWithError
+	var beforeFunc any
+
+	beforeFunc = j.beforeJobRuns
+
+	if j.beforeJobRunsWithError != nil {
```
JohnRoesler: I think `beforeJobRunsSkipIfBeforeFuncErrors` or something like that, describes what the function does better.

executor.go
```diff
@@ -389,7 +389,27 @@ func (e *executor) runJob(j internalJob, jIn jobIn) {
...
+		beforeFunc = j.beforeJobRunsWithError
+	}
```
JohnRoesler: I don't see a reason that this needs to override the existing `beforeJobRuns`. I see the following scenarios as all valid:
```
1. Not running either beforeJobRuns and beforeJobRunsSkipIfBeforeFuncErrors
2. Running either beforeJobRuns or beforeJobRunsSkipIfBeforeFuncErrors
3. Running both beforeJobRuns and beforeJobRunsSkipIfBeforeFuncErrors
```

I would have it run the `beforeJobRuns` func first. Note that the function `callJobFuncWithParams` doesn't do anything if the passed function is nil, which it will be if unset. Then run the `beforeJobRunsSkipIfBeforeFuncErrors` and do the logic you have below to reschedule and return if an error is returned.

job.go
```diff
@@ -724,6 +725,19 @@ func BeforeJobRuns(eventListenerFunc func(jobID uuid.UUID, jobName string)) Even
 	}
 }
 
+// BeforeJobRunsError is used to listen for when a job is about to run and
+// then run the provided function and returns an error if any.
+// This function overwrites BeforeJobRuns.
+func BeforeJobRunsError(eventListenerFunc func(jobID uuid.UUID, jobName string) error) EventListener {
```
```suggestion
// BeforeJobRunsSkipIfBeforeFuncErrors is used to listen for when a job is about to run and
// then runs the provided function. If the provided function returns an error, the job will be
// rescheduled and the current run will be skipped.
func BeforeJobRunsSkipIfBeforeFuncErrors(eventListenerFunc func(jobID uuid.UUID, jobName string) error) EventListener {
```

.github/workflows/file_formatting.yml
```diff
@@ -16,4 +16,4 @@ jobs:
         uses: actions/checkout@v4
       - name: verify example_test.go
         run: |
-          grep "^func " example_test.go | sort -c
+          grep "^func [a-z-A-Z]" example_test.go | sort -c
```
JohnRoesler, line 19: This is erroring in the actions run:
`grep: Invalid range end`
Should be
```suggestion
          grep "^func [a-zA-Z]" example_test.go | sort -c
```

JohnRoesler: Would you also update that in the Makefile?

executor.go
FalcoSuessgott: neat! even better - will incorporate

JohnRoesler (commented):
Last thing, can you add an example to the example test file?

JohnRoesler (approved):
Nice!

PR #819, opened 2025-01-23
cancel running jobs if possible when run Scheduler.StopJobs()

### LINKED ISSUE #818: [BUG] - A long running job can not be cancelled when run s.StopJobs()
### Describe the bug

A long running job can not be cancelled when run s.StopJobs(), and always return "timed out waiting for jobs to finish"

### To Reproduce

1. Create a Scheduler and add a long running job and set WithStartImmediately
2. start the scheduler
3. wait for a while before the job is completed, run s.StopJobs()
4. timeout error returns

### Version

v2.14.2

### Expected behavior

if the job has a ctx field, Scheduler should cancel it when running StopJobs() instead of waiting for it

27149chen:
### What does this do?

when run Scheduler.StopJobs(), the scheduler will wait for the running jobs to be completed. If there are long running jobs, it will always return "timed out waiting for jobs to finish".

This pr try to cancel the running jobs if possible

Resolves #818

You can now add a task function which accept ctx as the first param, but you do need pass a ctx actually, gocron will handle it properly
```
	j, err := s.NewJob(
		gocron.DurationJob(
			10*time.Second,
		),
		gocron.NewTask(
			func(ctx context.Context) {
				// do things
			},
		),
	)
```
but if you have you own ctx, fell free to pass it
```
	j, err := s.NewJob(
		gocron.DurationJob(
			10*time.Second,
		),
		gocron.NewTask(
			func(ctx context.Context) {
				// do things
			},
                         ctx,
		),
	)
```
or you can also pass it as a job option to control the whole thing
```
	j, err := s.NewJob(
		gocron.DurationJob(
			10*time.Second,
		),
		gocron.NewTask(
			func(ctx context.Context) {
				// do things
			},
		),
                gocron.WithContext(ctx)
	)
```

yes

- [ ] Updated `README.md`

### DIFF
--- a/scheduler_test.go
+++ b/scheduler_test.go
@@ -361,6 +361,42 @@ func TestScheduler_StopTimeout(t *testing.T) {
 	}
 }
 
+func TestScheduler_StopLongRunningJobs(t *testing.T) {
+	t.Run("start, run job, stop jobs before job is completed", func(t *testing.T) {
+		s := newTestScheduler(t,
+			WithStopTimeout(50*time.Millisecond),
+		)
+
+		_, err := s.NewJob(
+			DurationJob(
+				50*time.Millisecond,
+			),
+			NewTask(
+				func(ctx context.Context) {
+					select {
+					case <-ctx.Done():
+						time.Sleep(time.Second)
+					case <-time.After(100 * time.Millisecond):
+						t.Fatal("job can not been canceled")
+					}
+				},
+			),
+			WithStartAt(
+				WithStartImmediately(),
+			),
+			WithSingletonMode(LimitModeReschedule),
+		)
+		require.NoError(t, err)
+
+		s.Start()
+
+		time.Sleep(20 * time.Millisecond)
+		// the running job is canceled, no unexpected timeout error
+		require.NoError(t, s.StopJobs())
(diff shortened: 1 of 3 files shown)


scheduler_test.go
```diff
@@ -361,6 +361,42 @@ func TestScheduler_StopTimeout(t *testing.T) {
...
+
+		_, err := s.NewJob(
+			DurationJob(
+				50*time.Millisecond,
+			),
+			NewTask(
+				func(ctx context.Context) {
+					select {
+					case <-ctx.Done():
+						time.Sleep(time.Second)
```
JohnRoesler: This test is failing, which makes sense since on cancellation the job is sleeping here, which would prevent it from stopping.

Is there an intention I'm missing here?
```suggestion
```

JohnRoesler (approved):
Hey @27149chen this is a great addition. Some things that will be needed still:
- addition of the new method to `example_test.go`
- additional comments on the `NewJob` and `NewTask` functions explaining that if your task accepts a context as it's first parameter, gocron will pass in a context you can listen to for cancellation. Also, updating the `example_test.go` for the NewTask

I will make a PR to address.

PR #823, opened 2025-01-31
replace "golang.org/x/exp" with standard libraries

apocelipes:
### What does this do?

Since gocron requires go1.21.0+, we can replace "golang.org/x/exp/maps" and "golang.org/x/exp/slices" with the standard libraries "maps" and "slices". BTW, "golang.org/x/exp/slices" now is only a wrapper of "slice".

### Notes

"removeSliceDuplicatesInt" could also get some performance benifits:

```go
func BenchmarkRemoveSliceDuplicatesInt(b *testing.B) {
	for range b.N {
		removeSliceDuplicatesInt([]int{1, 4, 5, 4, 6, 2, 3})
	}
}
```

```text
goos: windows
goarch: amd64
pkg: bench
cpu: Intel(R) Core(TM) i5-10200H CPU @ 2.40GHz
                           │     old      │                 new                 │
                           │    sec/op    │   sec/op     vs base                │
RemoveSliceDuplicatesInt-8   285.00n ± 1%   29.75n ± 0%  -89.56% (p=0.000 n=10)

                           │    old     │                new                 │
                           │    B/op    │   B/op     vs base                 │
RemoveSliceDuplicatesInt-8   96.00 ± 0%   0.00 ± 0%  -100.00% (p=0.000 n=10)

                           │    old     │                 new                 │
                           │ allocs/op  │ allocs/op   vs base                 │
RemoveSliceDuplicatesInt-8   3.000 ± 0%   0.000 ± 0%  -100.00% (p=0.000 n=10)
```

As an added benefit, there is now no need to sort the results of “removeSliceDuplicatesInt” because they are already sorted.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/stretchr/testify v1.10.0
 	go.uber.org/goleak v1.3.0
-	golang.org/x/exp v0.0.0-20240808152545-0cdaa3abc0fa
 )
 
 require (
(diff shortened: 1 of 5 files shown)


JohnRoesler (approved):
Thanks for taking the time to do this!

PR #829, opened 2025-02-13
removes nextRuns in the past when job skipped by locker #828

### LINKED ISSUE #828: [BUG] - `NextRun()` method returning a value in the past when using a locker.
### Describe the bug

When I call the method `NextRun()` in a job I get a timestamp in the past. I think it's related to the fact that I am using a `gocron.Locker` and that particular instance does not get to do the job.
Then I can have the scenario that if for example, if my project have two instances, and I am exposing the `nextRun` value through an endpoint, the values are different for each instance.

I think I am not explaining myself very good, so that's why I created a repo to show it live, link below.

### To Reproduce

I created a gist to describe the scenario:
https://gist.github.com/manuelarte/e33cec46cd8653a5e2731687041a051a

### Version

github.com/go-co-op/gocron/v2 v2.15.0

### Expected behavior

I would expect the NextRun value to be the same independently if that particular instance got the job last time or not.

### Additional context

As mentioned before, I created a repo to easily show the issue https://github.com/manuelarte/cron-job-next-run-issue

manuelarte:
### What does this do?

It proposes a possible solution for the bug described in #828 

Resolves #828 

no breaking of current functionality

Will include them after we agree on the solution

- [ ] Updated `README.md`

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -31,6 +31,9 @@ type executor struct {
 	// used to request jobs from the scheduler
 	jobOutRequest chan jobOutRequest
 
+	// sends out job needs to update the next runs
+	jobUpdateNextRuns chan uuid.UUID
+
 	// used by the executor to receive a stop signal from the scheduler
 	stopCh chan struct{}
 	// the timeout value when stopping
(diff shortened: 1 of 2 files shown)


manuelarte:
Some somments:

I added a new channel to "update" the `nextScheduled` array. In the current draft I decided to create a new channel and a new method to handle this update of the `nextScheduled`, but there are other places where the `nextScheduled` array is updated and I am not reusing code.

I decided to create a new channel, but maybe it makes sense to modify the method `selectExecJobsOutForRescheduling` and also "remove" the old entries in `nextScheduled`. 

Or any other solution that I don't see right now.

What do you guys think?

JohnRoesler:
hey @manuelarte 

I considered a few options, and without changing any other behavior, this seems like a reasonable route to go. Thanks for digging in and putting forth a fix.

JohnRoesler (approved):

PR #834, opened 2025-02-20
feat:custom-cron interface for own custom cron implimentation

### LINKED ISSUE #804: [FEATURE] - run a job on the last weekday of the month
### Describe the bug

Gocron cronjob cant parse "LW", errors with invalid syntax

### To Reproduce

`
package main
import (
	"log"
)
func main() {
	s, err := gocron.NewScheduler()
	if err != nil {
		log.Fatal(err)
	}

	_, err = s.NewJob(
		gocron.CronJob("0 0 5 LW * ?", true),
		gocron.NewTask(func() {
			log.Println("Test job 1")
		}),
	)
	if err != nil {
		log.Printf("Test job 1 err: %v", err)
	}

	_, err = s.NewJob(
		gocron.CronJob("0 5 LW * ?", false),
		gocron.NewTask(func() {
			log.Println("Test job 2")
		}),
	)
	if err != nil {
		log.Printf("Test job 2 err: %v", err)
	}

	s.Start()
	for {
	}
}
`

### Version

v2.13.0

### Expected behavior

The expression should be parsed since its a valid cron expression

Dojeto:
### What does this do?
This pr for the new feature of custom cron job, added the cron interface for custom cron implementation 

Resolves #804

Errors will be change since from two different error `ErrCronJobParse` and `ErrCronJobInvalid` i have change it to `ErrCronJobInvalid` only one which leads to failure in test cases.

No

No

- [⛌] Updated `example_test.go`
- [⛌] Updated `README.md`

### Notes
This is very straight forward approach i have think this needs more refinements, suggestions appreciated

### DIFF
--- a/job.go
+++ b/job.go
@@ -115,8 +115,8 @@ type limitRunsTo struct {
 // implemented to provide a custom cron implementation for
 // the job. Pass in the implementation using the JobOption WithCronImplementation.
 type Cron interface {
-	IsValid(string) error
-	Next(string, time.Time) time.Time
+	IsValid(crontab string, location *time.Location, now time.Time) error
+	Next(lastRun time.Time) time.Time
 }
 
 // -----------------------------------------------
--- a/job.go
+++ b/job.go
@@ -107,15 +107,15 @@ type limitRunsTo struct {
 
 // -----------------------------------------------
 // -----------------------------------------------
-// --------------- Custom Cron ------------------
+// --------------- Custom Cron -------------------
 // -----------------------------------------------
 // -----------------------------------------------
 
 // Cron defines the interface that must be
-// implemented to create a cron.
-
+// implemented to provide a custom cron implementation for
+// the job. Pass in the implementation using the JobOption WithCronImplementation.
 type Cron interface {
-	IsValid(string) bool
+	IsValid(string) error
 	Next(string, time.Time) time.Time
 }
 
@@ -131,20 +131,20 @@ type JobDefinition interface {
 	setup(j *internalJob, l *time.Location, now time.Time) error
 }
 
-// Default cron implementation using robfig
+// Default cron implementation
 
 func newDefaultCronImplementation(withSeconds bool) Cron {
-	return &RobfigCron{
+	return &defaultCron{
 		withSeconds: withSeconds,
 	}
 }
 
-type RobfigCron struct {
+type defaultCron struct {
 	cronSchedule cron.Schedule
 	withSeconds  bool
 }
 
-func (r *RobfigCron) IsValid(crontab string) bool {
+func (r *defaultCron) IsValid(crontab string) error {
 	var withLocation string
 	if strings.HasPrefix(crontab, "TZ=") || strings.HasPrefix(crontab, "CRON_TZ=") {
 		withLocation = crontab
@@ -166,16 +166,16 @@ func (r *RobfigCron) IsValid(crontab string) bool {
 		cronSchedule, err = cron.ParseStandard(withLocation)
 	}
 	if err != nil {
-		return false
+		return errors.Join(ErrCronJobParse, err)
 	}
 	if cronSchedule.Next(time.Now()).IsZero() {
-		return false
+		return ErrCronJobInvalid
 	}
 	r.cronSchedule = cronSchedule
-	return true
+	return nil
 }
 
-func (r *RobfigCron) Next(crontab string, lastRun time.Time) time.Time {
+func (r *defaultCron) Next(crontab string, lastRun time.Time) time.Time {
 	return r.cronSchedule.Next(lastRun)
 }
 
@@ -192,7 +192,8 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, _ time
 		c.cron = j.cron
 	}
 
-	if !c.cron.IsValid(c.crontab) {
+	if err := c.cron.IsValid(c.crontab); err != nil {
+	    return err
 		return ErrCronJobInvalid
 	}
 
@@ -653,7 +654,8 @@ func WithName(name string) JobOption {
 	}
 }
 
-// JobOption to set custom Cron implementation
+// WithCronImplementation sets the custom Cron implementation for the job.
+// This is only utilized for the CronJob type.
 func WithCronImplementation(c Cron) JobOption {
 	return func(j *internalJob, _ time.Time) error {
 		j.cron = c
--- a/job.go
+++ b/job.go
@@ -106,6 +106,20 @@ type limitRunsTo struct {
 
 // -----------------------------------------------
 // -----------------------------------------------
+// --------------- Custom Cron ------------------
+// -----------------------------------------------
+// -----------------------------------------------
+
+// Cron defines the interface that must be
+// implemented to create a cron.
+
+type Cron interface {
+	IsValid() bool
+	Next(time.Time) time.Time
+}
+
+// -----------------------------------------------
+// -----------------------------------------------
 // --------------- Job Variants ------------------
 // -----------------------------------------------
 // -----------------------------------------------
@@ -116,21 +130,45 @@ type JobDefinition interface {
 	setup(j *internalJob, l *time.Location, now time.Time) error
 }
 
+var _ JobDefinition = (*customCron)(nil)
+
+type customCron struct {
+	crontab string
+	cron    Cron
+}
+
+func (c customCron) setup(j *internalJob, location *time.Location, now time.Time) error {
+	if !c.cron.IsValid() {
+		return ErrCronJobInvalid
+	}
+	j.jobSchedule = &cronJob{cronSchedule: c.cron}
+	return nil
+}
+
+func CustomCronJob(crontab string, cron Cron) JobDefinition {
+	return customCron{
+		crontab: crontab,
+		cron:    cron,
+	}
+}
+
+// default cron job implimentation
 var _ JobDefinition = (*cronJobDefinition)(nil)
 
 type cronJobDefinition struct {
-	crontab     string
-	withSeconds bool
+	crontab      string
+	withSeconds  bool
+	cronSchedule cron.Schedule
 }
 
-func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
+func (c *cronJobDefinition) IsValid() bool {
 	var withLocation string
 	if strings.HasPrefix(c.crontab, "TZ=") || strings.HasPrefix(c.crontab, "CRON_TZ=") {
 		withLocation = c.crontab
 	} else {
 		// since the user didn't provide a timezone default to the location
 		// passed in by the scheduler. Default: time.Local
-		withLocation = fmt.Sprintf("CRON_TZ=%s %s", location.String(), c.crontab)
+		withLocation = fmt.Sprintf("CRON_TZ=%s %s", time.Local, c.crontab)
 	}
 
 	var (
@@ -145,13 +183,24 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
 		cronSchedule, err = cron.ParseStandard(withLocation)
 	}
 	if err != nil {
-		return errors.Join(ErrCronJobParse, err)
+		return false
 	}
-	if cronSchedule.Next(now).IsZero() {
+	if cronSchedule.Next(time.Now()).IsZero() {
+		return false
+	}
+	c.cronSchedule = cronSchedule
+	return true
+}
+
+func (c *cronJobDefinition) Next(lastRun time.Time) time.Time {
+	return c.cronSchedule.Next(lastRun)
+}
+
+func (c *cronJobDefinition) setup(j *internalJob, location *time.Location, _ time.Time) error {
+	if !c.IsValid() {
 		return ErrCronJobInvalid
 	}
-
-	j.jobSchedule = &cronJob{cronSchedule: cronSchedule}
+	j.jobSchedule = &cronJob{cronSchedule: c}
 	return nil
 }
 
@@ -162,7 +211,7 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
 // crontab in the form `TZ=America/Chicago * * * * *` or
 // `CRON_TZ=America/Chicago * * * * *`
 func CronJob(crontab string, withSeconds bool) JobDefinition {
-	return cronJobDefinition{
+	return &cronJobDefinition{
 		crontab:     crontab,
 		withSeconds: withSeconds,
 	}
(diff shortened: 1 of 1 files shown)


Dojeto:
@JohnRoesler what's your thoughts on this implementation, my test cases are getting failed due to combine the two different error into one let me know what kinda changes require in this implementation and what should we do for the error ? if you allow me to do with one error then i will change testcases and test and refine it

job.go
```diff
@@ -116,21 +130,45 @@ type JobDefinition interface {
...
 type cronJobDefinition struct {
-	crontab     string
-	withSeconds bool
+	crontab      string
+	withSeconds  bool
+	cronSchedule cron.Schedule
 }
 
-func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
+func (c *cronJobDefinition) IsValid() bool {
```
JohnRoesler: rather than having the JobDefinition implement the Cron interface, I was envisioning that the cronJobDefinition itself would include the Cron interface as a struct field (it will also need to be added to the internalJob struct). Then, in the `setup` method of the `cronJobDefinition`, could grab the Cron interface from the internalJob and utilize that.

The default Cron implementation would be the current one with robfig. Then, add a new `JobOption` that takes in a Cron interface and sets it.

JohnRoesler: This solution keeps us to a single CronJob definition and instead allows altering the Cron implementation via a Job option which can be set for a single job, or globally using the `WithGlobalJobOptions` method on the scheduler.

Dojeto: There are few points that i want to discuss :

1. Do i have to make a different Jobdefination function since it has extra field which might cause confusion on the Custom implementation via Job option !

2. Where we are going to initialized the cron ?

`func WithCronImplementation(c Cron) JobOption {
	return func(j *internalJob, _ time.Time) error {
		j.cron = c
		return nil
	}
}`

`func (c cronJobDefinition) setup(j *internalJob, location *time.Location, _ time.Time) error {
	c.cron = j.cron`
	
i hope this is exactly what you are trying to say

i am confused like initialize job and passing implementation is different thing Job initializing happening in `job definition ` and implementation passing through `job  options` how we can use them together. sorry if i am asking dumb question this pattern is new to me and i like this pattern :)

JohnRoesler: @Dojeto No worries on asking questions!

1. No, I don't think you need a different JobDefinition. if the implementation doesn't care about the `withSeconds` bool it can simply ignore it!
2. Yeah, a little confusing maybe, I'll try and explain:

The Job itself is "initialized" in the [NewJob](https://github.com/go-co-op/gocron/blob/v2/scheduler.go#L602) method on the Scheduler, which in turn just calls the internal method [addOrUpdateJob](https://github.com/go-co-op/gocron/blob/v2/scheduler.go#L670). If you look in that method, you'll see the following order of operations:
1. Job options are applied, first global, then job specific so that they win -> [here](https://github.com/go-co-op/gocron/blob/v2/scheduler.go#L706-L718) 
2. Then the jobDefinitions `setup` method is called https://github.com/go-co-op/gocron/blob/v2/scheduler.go#L742

So, in this case, the Cron interface will be initialized either by the caller and then passed in via the `WithCronImplementation` JobOption, which in turn sets the Cron onto the internalJob struct. OR, the Cron will be initialized inside the `setup` method as the default implementation.

Then, in the `setup` method, `IsValid` should be called.

Dojeto:
@JohnRoesler how it is ? test cases getting failed due to change in error, i have check seems like it is working, let me know what's your thoughts on this ? let me know required changes.

job.go
```diff
@@ -104,6 +105,20 @@ type limitRunsTo struct {
...
+// -----------------------------------------------
+// --------------- Custom Cron ------------------
+// -----------------------------------------------
+// -----------------------------------------------
+
+// Cron defines the interface that must be
+// implemented to create a cron.
+
+type Cron interface {
+	IsValid(string) bool
```
JohnRoesler, line 118: let's have this return an error instead of a bool

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
 	setup(j *internalJob, l *time.Location, now time.Time) error
 }
 
-var _ JobDefinition = (*cronJobDefinition)(nil)
+// Default cron implementation using robfig
```
```suggestion
// Default cron implementation
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
-	crontab     string
-	withSeconds bool
+func newDefaultCronImplementation(withSeconds bool) Cron {
+	return &RobfigCron{
+		withSeconds: withSeconds,
+	}
 }
 
-func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
+type RobfigCron struct {
```
```suggestion
type defaultCron struct {
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
-	if c.withSeconds {
+	if r.withSeconds {
 		p := cron.NewParser(cron.SecondOptional | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor)
 		cronSchedule, err = p.Parse(withLocation)
 	} else {
 		cronSchedule, err = cron.ParseStandard(withLocation)
 	}
 	if err != nil {
-		return errors.Join(ErrCronJobParse, err)
+		return false
```
```suggestion
		return errors.Join(ErrCronJobParse, err)
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
+	}
 }
 
-func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
+type RobfigCron struct {
+	cronSchedule cron.Schedule
+	withSeconds  bool
+}
+
+func (r *RobfigCron) IsValid(crontab string) bool {
```
JohnRoesler, line 147:
```suggestion
func (r *RobfigCron) IsValid(crontab string) error {
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
 		cronSchedule, err = p.Parse(withLocation)
 	} else {
 		cronSchedule, err = cron.ParseStandard(withLocation)
 	}
 	if err != nil {
-		return errors.Join(ErrCronJobParse, err)
+		return false
+	}
+	if cronSchedule.Next(time.Now()).IsZero() {
+		return false
```
```suggestion
		return ErrCronJobInvalid
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
 	}
-	if cronSchedule.Next(now).IsZero() {
+	r.cronSchedule = cronSchedule
+	return true
```
```suggestion
	return nil
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
+	crontab string
+	cron    Cron
+}
+
+func (c cronJobDefinition) setup(j *internalJob, location *time.Location, _ time.Time) error {
+	if j.cron != nil {
+		c.cron = j.cron
+	}
+
+	if !c.cron.IsValid(c.crontab) {
```
```suggestion
	if err := c.cron.IsValid(c.crontab); err != nil {
	    return err
```

job.go
```diff
@@ -104,6 +105,20 @@ type limitRunsTo struct {
 	runCount uint
 }
 
+// -----------------------------------------------
+// -----------------------------------------------
+// --------------- Custom Cron ------------------
+// -----------------------------------------------
```
```suggestion
// -----------------------------------------------
// --------------- Custom Cron -------------------
// -----------------------------------------------
```

job.go
```diff
@@ -104,6 +105,20 @@ type limitRunsTo struct {
...
+// -----------------------------------------------
+
+// Cron defines the interface that must be
+// implemented to create a cron.
+
+type Cron interface {
```
```suggestion
// Cron defines the interface that must be
// implemented to provide a custom cron implementation for
// the job. Pass in the implementation using the JobOption WithCronImplementation.
type Cron interface {
```

job.go
```suggestion
	IsValid(string) error
```

job.go
```suggestion
func (r *defaultCron) IsValid(crontab string) error {
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
+	}
+	if cronSchedule.Next(time.Now()).IsZero() {
+		return false
 	}
-	if cronSchedule.Next(now).IsZero() {
+	r.cronSchedule = cronSchedule
+	return true
+}
+
+func (r *RobfigCron) Next(crontab string, lastRun time.Time) time.Time {
```
```suggestion
func (r *defaultCron) Next(crontab string, lastRun time.Time) time.Time {
```

job.go
```diff
@@ -608,6 +653,14 @@ func WithName(name string) JobOption {
 	}
 }
 
+// JobOption to set custom Cron implementation
```
```suggestion
// WithCronImplementation sets the custom Cron implementation for the job.
// This is only utilized for the CronJob type.
```

job.go
```diff
@@ -116,42 +131,72 @@ type JobDefinition interface {
...
 }
 
-var _ JobDefinition = (*cronJobDefinition)(nil)
+// Default cron implementation using robfig
 
-type cronJobDefinition struct {
-	crontab     string
-	withSeconds bool
+func newDefaultCronImplementation(withSeconds bool) Cron {
+	return &RobfigCron{
```
```suggestion
	return &defaultCron{
```

job.go
```diff
@@ -147,11 +168,36 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
...
 	}
-	if cronSchedule.Next(now).IsZero() {
+	if cronSchedule.Next(time.Now()).IsZero() {
 		return ErrCronJobInvalid
 	}
+	r.cronSchedule = cronSchedule
+	return nil
+}
+
+func (r *defaultCron) Next(crontab string, lastRun time.Time) time.Time {
```
```suggestion
func (r *defaultCron) Next(lastRun time.Time) time.Time {
```

job.go
```diff
@@ -116,29 +131,35 @@ type JobDefinition interface {
...
-type cronJobDefinition struct {
-	crontab     string
-	withSeconds bool
+func newDefaultCronImplementation(withSeconds bool) Cron {
+	return &defaultCron{
+		withSeconds: withSeconds,
+	}
+}
+
+type defaultCron struct {
```
JohnRoesler, line 144:
```suggestion
var _ Cron = (*defaultCron)(nil)

type defaultCron struct {
```

job.go
```diff
@@ -104,6 +105,20 @@ type limitRunsTo struct {
...
+// -----------------------------------------------
+// --------------- Custom Cron -------------------
+// -----------------------------------------------
+// -----------------------------------------------
+
+// Cron defines the interface that must be
+// implemented to provide a custom cron implementation for
+// the job. Pass in the implementation using the JobOption WithCronImplementation.
+type Cron interface {
+	IsValid(string) error
```
```suggestion
	IsValid(crontab string, location *time.Location, now time.Time) error
```

job.go
```diff
@@ -116,29 +131,35 @@ type JobDefinition interface {
...
+	}
+}
+
+type defaultCron struct {
+	cronSchedule cron.Schedule
+	withSeconds  bool
 }
 
-func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
+func (r *defaultCron) IsValid(crontab string) error {
```
```suggestion
func (r *defaultCron) IsValid(crontab string, location *time.Location, now time.Time) error {
```

job.go
```diff
@@ -116,29 +131,35 @@ type JobDefinition interface {
...
 	var withLocation string
-	if strings.HasPrefix(c.crontab, "TZ=") || strings.HasPrefix(c.crontab, "CRON_TZ=") {
-		withLocation = c.crontab
+	if strings.HasPrefix(crontab, "TZ=") || strings.HasPrefix(crontab, "CRON_TZ=") {
+		withLocation = crontab
 	} else {
 		// since the user didn't provide a timezone default to the location
 		// passed in by the scheduler. Default: time.Local
-		withLocation = fmt.Sprintf("CRON_TZ=%s %s", location.String(), c.crontab)
+		withLocation = fmt.Sprintf("CRON_TZ=%s %s", time.Local, crontab)
```
```suggestion
		withLocation = fmt.Sprintf("CRON_TZ=%s %s", location.String(), crontab)
```

job.go
```diff
@@ -147,11 +168,36 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
 	if err != nil {
 		return errors.Join(ErrCronJobParse, err)
 	}
-	if cronSchedule.Next(now).IsZero() {
+	if cronSchedule.Next(time.Now()).IsZero() {
```
```suggestion
	if cronSchedule.Next(now).IsZero() {
```

job.go
```diff
@@ -104,6 +105,20 @@ type limitRunsTo struct {
...
+// --------------- Custom Cron -------------------
+// -----------------------------------------------
+// -----------------------------------------------
+
+// Cron defines the interface that must be
+// implemented to provide a custom cron implementation for
+// the job. Pass in the implementation using the JobOption WithCronImplementation.
+type Cron interface {
+	IsValid(string) error
+	Next(string, time.Time) time.Time
```
```suggestion
	Next(lastRun time.Time) time.Time
```

job.go
```diff
@@ -147,11 +168,36 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
...
-	j.jobSchedule = &cronJob{cronSchedule: cronSchedule}
+// default cron job implimentation
+var _ JobDefinition = (*cronJobDefinition)(nil)
+
+type cronJobDefinition struct {
+	crontab string
+	cron    Cron
+}
+
+func (c cronJobDefinition) setup(j *internalJob, location *time.Location, _ time.Time) error {
```
```suggestion
func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now time.Time) error {
```

job.go
```diff
@@ -147,11 +168,36 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
...
+	if j.cron != nil {
+		c.cron = j.cron
+	}
+
+	if err := c.cron.IsValid(c.crontab); err != nil {
```
```suggestion
	if err := c.cron.IsValid(c.crontab, location, now); err != nil {
```

job.go
```diff
@@ -147,11 +168,36 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
...
+	    return err
+		return ErrCronJobInvalid
```
```suggestion
```

job.go
```diff
@@ -818,11 +874,12 @@ type jobSchedule interface {
...
 
 type cronJob struct {
-	cronSchedule cron.Schedule
+	crontab      string
+	cronSchedule Cron
 }
 
 func (j *cronJob) next(lastRun time.Time) time.Time {
-	return j.cronSchedule.Next(lastRun)
+	return j.cronSchedule.Next(j.crontab, lastRun)
```
```suggestion
	return j.cronSchedule.Next(lastRun)
```

JohnRoesler (approved):

PR #836, opened 2025-02-26
Fix #835 and #837

### LINKED ISSUE #835: [BUG] - Executor deadlocks when it calls `stop()`.
### Describe the bug

Image

Sometimes the unit tests could block for 10 mins, this is not normal.

### To Reproduce

Run the unit test a few more times and there's a chance you'll see the deadlock.

I read the log and the traceback, then I found the process was blocked by this:

```golang
func (e *executor) stop(standardJobsWg, singletonJobsWg, limitModeJobsWg *waitGroupWithMutex) {
	...
	if count < 3 {
		e.done <- ErrStopJobsTimedOut
		e.logger.Debug("gocron: executor stopped - timed out")
	} else {
		e.done <- nil  // blocking
		e.logger.Debug("gocron: executor stopped")
	}
	waiterCancel()

	if e.limitMode != nil {
		e.limitMode.started = false
	}
}
```

One possible reason is that waiting for all jobs to complete takes so much time that the scheduler thinks the exit timeout (although the default timeout now is 11 seconds, it is still possible to timeout) is over and stops receiving data from “e.done”, causing the sender to block permanently.

### Expected behavior

No deadlock, no timeout.

### Additional context

There are two ways to fix this deadlock: increase the scheduler wait timeout or turn “e.done” into a buffered chan. IMHO, changing the timeout does not fundamentally solve the problem.

### LINKED ISSUE #837: [BUG] - Goroutine leaks when shutdown
### Describe the bug

Image

### To Reproduce

Appears randomly in unit tests.

One possible reason is that “s.exec.stopCh” has multiple receivers, but only sends the signal once, causing the receiver that doesn't get the signal to block permanently.

### Expected behavior

No leaks.

apocelipes:
### What does this do?

Fixes #835.
Fixes #837.

For 837: `e.stopCh` now has only one receiver, and if `executorCtx.Done()` does not block, it means that the executor has already received the stop signal and called the stop method, so there is no need to do additional operations, just return from the sub-goroutine.
I also use `sync.Once` to prevent the stop method from being called multiple times before the next call to the start method. This may not be necessary, but it maximizes concurrency security, and it doesn't do any noticeable harm.

#835 
#837 

- [ ] Updated `README.md`

### Notes

The two issues have some relevance, so I put the fixes in the same PR. Please don't sqush the commits.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -36,6 +36,8 @@ type executor struct {
 
 	// used by the executor to receive a stop signal from the scheduler
 	stopCh chan struct{}
+	// ensure that stop runs before the next call to start and only runs once
+	stopOnce *sync.Once
 	// the timeout value when stopping
 	stopTimeout time.Duration
 	// used to signal that the executor has completed shutdown
@@ -88,6 +90,7 @@ func (e *executor) start() {
 	// any other uses within the executor should create a context
 	// using the executor context as parent.
 	e.ctx, e.cancel = context.WithCancel(context.Background())
+	e.stopOnce = &sync.Once{}
 
 	// the standardJobsWg tracks
 	standardJobsWg := &waitGroupWithMutex{}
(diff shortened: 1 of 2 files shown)


JohnRoesler (approved):
Thanks for digging into these. Concurrency is hard!

PR #850, opened 2025-05-28
opt: output detailed stack trace information when a panic occurs.

coticom:


JohnRoesler:
hi @coticom rather than putting this directly in, as not all users may desire this verbose output, does setting up the panic handler to log the stack meet your needs?

```go
package main

import (
	"log"
	"runtime/debug"
	"time"

	"github.com/google/uuid"

)

func main() {
	s, err := gocron.NewScheduler(
		gocron.WithGlobalJobOptions(
			gocron.WithEventListeners(
				gocron.AfterJobRunsWithPanic(
					func(jobID uuid.UUID, jobName string, recoverData any) {
						log.Printf("Job %s with ID %s panicked with data: %v\n", jobName, jobID, recoverData)
						log.Printf("stack: %s\n", string(debug.Stack()))
					},
				),
			),
		),
	)
	if err != nil {
		log.Fatal(err)
	}

	_, err = s.NewJob(
		gocron.DurationJob(time.Second),
		gocron.NewTask(func() {
			log.Println("job ran")
			panic("this is a test panic to see if the job will recover and continue running after a panic")
		}),
	)
	if err != nil {
		log.Fatal(err)
	}

	s.Start()

	select {}
}
```

PR #858, opened 2025-07-21
fix: cancel job contexts in create/update errors

### LINKED ISSUE #857: [BUG] - Leaked context: context.WithCancel not cleaned up on job setup failure
### Describe the bug

The scheduler’s addOrUpdateJob function creates a new context using context.WithCancel, but in some failure scenarios (e.g., if definition.setup(...) returns an error), the associated cancel() function is never called. This causes the context and its cancel chain to remain in memory, resulting in a memory leak over time.

Additionally, if the job is replaced or removed, but its cancel() function is not triggered via worker logic or explicit cleanup, the memory held by the context continues to grow.

### To Reproduce

1.	Call addOrUpdateJob(...) with a JobDefinition that fails during setup (eg., the scheduled time is in the past).
2.	Observe that context.WithCancel(...) is called, but cancel() is never invoked.
3.	Repeat the above many times or with a scheduler that retries tasks.
4.	Use pprof to inspect heap: note increasing number of context.cancelCtx or related allocations.
5.	Optionally run with GODEBUG=gcshrinkstackoff=1 and pprof/trace to see retained context trees.

# Heap observation

Image

### Version

github.com/go-co-op/gocron/v2 v2.13.0

### Expected behavior

•	The context created with context.WithCancel should be properly canceled if job creation/setup fails.
•	Job context should not be retained unnecessarily if the job was never started or was replaced.

### Additional context

N/A

JohnRoesler:
### What does this do?
cancel the job context, and new job context in the event that job creation errors or we get caught in scheduler shutdown 

resolves #857 

- [ ] Updated `README.md`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/stretchr/testify v1.10.0
 	go.uber.org/goleak v1.3.0
+	golang.org/x/crypto v0.40.0
 )
 
 require (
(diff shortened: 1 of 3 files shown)


VoxT:
Could you please release a fixed version for this? @JohnRoesler

PR #870, opened 2025-09-02
Use `errors.New` for non-formatted strings

apocelipes:
### What does this do?

Use `errors.New` for non-formatted strings.

`errors.New` performs better, and since `fmt.Errorf` ultimately calls `errors.New` if there is no error needs be wrapped, we can use `errors.New` directly.

No functionality changes.

### Notes

Benchmarks:

```go
func BenchmarkFmtErrorf(b *testing.B) {
	for b.Loop() {
		_ = fmt.Errorf("gocron: DailyJob: atTimes minutes and seconds must be between 0 and 59 inclusive")
	}
}

func BenchmarkErrorsNew(b *testing.B) {
	for b.Loop() {
		_ = errors.New("gocron: DailyJob: atTimes minutes and seconds must be between 0 and 59 inclusive")
	}
}
```

image

`errors.New` is about 6 times faster and uses less memory.


JohnRoesler (approved):
Thanks for the detailed PR description!

PR #872, opened 2025-09-17
a return is missing to stop execution if the job is not found

sarff:
### What does this do?
Added a return to the selectRunJobRequest function

The existing tests remain, and they have passed


JohnRoesler (approved):

JohnRoesler:
@sarff thanks for catching this!

Issue #904, opened 2025-12-09
[BUG] - Job's NextRun time is inaccurate. The job has already run, but NextRun has not been updated to the next scheduled time.


ThinkingXuan:
### Describe the bug

Job's NextRun time is inaccurate. The job has already run, but NextRun has not been updated to the next scheduled time.

Multiple Jobs run into this situation when executing simultaneously.

``` go
// create a scheduler
	scheduler, err := gocron.NewScheduler()
	if err != nil {
		log.Fatal(err)
	}

	// add a job to the scheduler
	job1, err := scheduler.NewJob(
		gocron.DurationJob(1*time.Minute),
		gocron.NewTask(func() {
			log.Println("Job executed start")
			time.Sleep(10 * time.Second)
			log.Println("Job executed")
		}),
		gocron.WithName("example-job"),
		gocron.WithTags("important"),
	)

	// add a job to the scheduler
	job2, err := scheduler.NewJob(
		gocron.DurationJob(2*time.Minute),
		gocron.NewTask(func() {
			log.Println("Job executed start")
			time.Sleep(10 * time.Second)
			log.Println("Job executed")
		}),
		gocron.WithName("example-job"),
		gocron.WithTags("important"),
	)

	job3, err := scheduler.NewJob(
		gocron.DurationJob(3*time.Minute),
		gocron.NewTask(func() {
			log.Println("Job executed start")
			time.Sleep(10 * time.Second)
			log.Println("Job executed")
		}),
		gocron.WithName("example-job"),
		gocron.WithTags("important"),
	)

	job4, err := scheduler.NewJob(
		gocron.DurationJob(4*time.Minute),
		gocron.NewTask(func() {
			log.Println("Job executed start")
			time.Sleep(10 * time.Second)
			log.Println("Job executed")
		}),
		gocron.WithName("example-job"),
		gocron.WithTags("important"),
	)

	if err != nil {
		log.Fatal(err)
	}

	// start the scheduler
	scheduler.Start()
```

### Version
gocorn: v2.18.2
go: 1.24.10

iyashjayesh:
Thanks for the report! I tried reproducing this. Seems like the issue is that `nextScheduled` isn't being updated atomically when multiple jobs run concurrently, it's a race condition in the state management. I'll investigate the executor's job completion logic and get a fix out. 

@JohnRoesler Please correct if my understanding is incorrect

PR #907, opened 2026-01-28
[fix] race condition in `NextRun()` when multiple jobs complete simultaneously

barkhayot:
### What does this do?
When multiple jobs completed at the same time, `NextRun()` could return stale/incorrect values due to a race condition in the `nextScheduled` cleanup logic. The issue occurred because two different methods (`selectExecJobsOutCompleted` and `updateNextScheduled`) were cleaning up the `nextScheduled` slice with different logic:
- One used `!t.Before(now)` (keeps times >= now)
- One used `t.After(now)` (keeps times > now)

When these methods ran concurrently on different jobs, their writes could interleave and overwrite each other with inconsistent states.

### Solution
Unified both methods to use identical cleanup logic: `t.After(now)` (keep only strictly future times). This makes the cleanup idempotent - both methods now produce the same result regardless of execution order, eliminating the race condition.

- fixes issue #904 
- **scheduler.go**: Updated `selectExecJobsOutCompleted` and `updateNextScheduled` to use consistent `t.After(now)` logic for removing past scheduled times
- **job_test.go**: 
  - Added `TestJob_NextRun_MultipleJobsSimultaneously` to reproduce and verify the fix for the original bug
  - Added `TestJob_NextRun_ConcurrentCompletions` to stress-test concurrent NextRun() queries
  - Fixed `TestJob_NextRuns` weekly job assertion - with proper cleanup, all intervals are now consistently 14 days (no special case for first run needed)

### Testing
All existing tests pass. New regression tests verify `NextRun()` returns correct values when multiple jobs complete simultaneously.

### DIFF
--- a/job_test.go
+++ b/job_test.go
@@ -1295,3 +1295,91 @@ func TestJob_NextRun_MultipleJobsSimultaneously(t *testing.T) {
 
 	require.NoError(t, s.Shutdown())
 }
+
+func TestJob_NextRun_ConcurrentCompletions(t *testing.T) {
+	// This test verifies that when multiple jobs complete at exactly the same time,
+	// their NextRun() values are correctly updated without race conditions.
+
+	testTime := time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)
+	fakeClock := clockwork.NewFakeClockAt(testTime)
+
+	s := newTestScheduler(t,
+		WithClock(fakeClock),
+		WithLocation(time.UTC), // Set scheduler to use UTC to match our test time
+	)
+
+	var wg sync.WaitGroup
+	jobCompletionBarrier := make(chan struct{})
+
+	// Create jobs that will all complete at the same instant
+	createJob := func(name string, interval time.Duration) Job {
+		job, err := s.NewJob(
+			DurationJob(interval),
+			NewTask(func() {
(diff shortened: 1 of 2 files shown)


job_test.go
```diff
@@ -630,12 +630,10 @@ func TestJob_NextRuns(t *testing.T) {
 				),
 			),
 			func(t *testing.T, iteration int, previousRun, nextRun time.Time) {
+				// With the fix for NextRun accuracy, the immediate run (Jan 1) is removed
+				// from nextScheduled after it completes. So all intervals should be 14 days
+				// (2 weeks as configured).
 				diff := time.Hour * 14 * 24
-				if iteration == 1 {
```
JohnRoesler, line 634: Lint is complaining that the iteration value passed in is unused now. Looks like it can be removed entirely from the assertion func

JohnRoesler (commented):
This is great!

job_test.go
barkhayot: Thanks for catching! Made quick fix here https://github.com/go-co-op/gocron/pull/907/changes/0f55e2ad3505e41eb9c1f2f50fbbec0e81326b6c based on feedback 🙏

JohnRoesler (approved):
💯 thank you!

PR #914, opened 2026-04-02
perf: replace string allocation with bytes.Compare in Jobs() UUID sort

huynhanx03:
### What does this do?

`Jobs()` sorts results by UUID before returning. The comparator was calling `uuid.UUID.String()` on both sides, allocating two strings per comparison. Since `uuid.UUID` is `[16]byte`, we can compare raw bytes directly.

N/A

None.

## Benchmark
```
goos: darwin
goarch: arm64
cpu: Apple M1

name                    old time/op      new time/op    delta
SchedulerJobs/10-8       4745 ns/op       2848 ns/op     -40%
SchedulerJobs/100-8     67616 ns/op      19742 ns/op     -71%
SchedulerJobs/500-8    435554 ns/op     104137 ns/op     -76%

name                  old allocs/op    new allocs/op    delta
SchedulerJobs/10-8               71               12     -83%
SchedulerJobs/100-8            1431              102     -93%
SchedulerJobs/500-8            9723              502     -95%

name                       old B/op         new B/op    delta
SchedulerJobs/10-8           3887 B           1072 B     -72%
SchedulerJobs/100-8         73725 B           9904 B     -87%
SchedulerJobs/500-8        490939 B          48304 B     -90%
```

Existing `TestScheduler_Jobs/order_is_equal` covers sort correctness.
Added `BenchmarkSchedulerJobs` in `scheduler_test.go`.

No API change.

### DIFF
--- a/scheduler.go
+++ b/scheduler.go
@@ -2,12 +2,12 @@
 package gocron
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"reflect"
 	"runtime"
 	"slices"
-	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


huynhanx03:
Hi @JohnRoesler , could you review, thanks.

JohnRoesler (approved):

PR #922, opened 2026-04-19
fix: remove jobs and respect stopTime in NextRuns when WithStopDateTi…

jcheyer:
fix: remove jobs and respect stopTime in NextRuns when WithStopDateTime is set

  Jobs configured with WithStopDateTime were not being removed from the
  scheduler after their stop time elapsed — selectExecJobsOutForRescheduling
  returned early without calling selectRemoveJob, and selectNewJob /
  selectStart scheduled the next timer even when it fell past stopTime.

  Additionally, NextRuns() returned future run times beyond stopTime,
  making the returned slice inconsistent with actual scheduler behaviour.

  Changes:
  - Call selectRemoveJob in selectExecJobsOutForRescheduling when
    stopTimeReached(), and also when the computed next run is >= stopTime
  - Guard selectNewJob and selectStart with the same stopTime check so
    jobs are removed immediately if their first/next run is out of range
  - Truncate NextRuns() output at stopTime instead of always filling
    the requested count

  Fixes: NextRuns returning times past stopTime, jobs staying in the
  scheduler indefinitely after WithStopDateTime elapses.

### DIFF
--- a/job.go
+++ b/job.go
@@ -1635,7 +1635,11 @@ func (j job) NextRuns(count int) ([]time.Time, error) {
 		}
 
 		from := out[i-1]
-		out[i] = ij.next(from)
+		next := ij.next(from)
+		if !ij.stopTime.IsZero() && !next.Before(ij.stopTime) {
+			return out[:i], nil
+		}
+		out[i] = next
 	}
 
 	return out, nil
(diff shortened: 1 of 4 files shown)


JohnRoesler (approved):
nice, thanks for providing a fix to the bug you found 🥳

PR #937, opened 2026-07-08
fix: eliminate shared mutable state across jobs (C4, H5)

JohnRoesler:
Correctness pass on two shared-mutable-state hazards that were missed in earlier PRs. Both hardened by dedicated `-race` regression tests.

## C4 — `defaultCron.IsValid` no longer mutates the receiver

**Before:** `defaultCron.IsValid` parsed the crontab and cached the resulting `cron.Schedule` on `c.cronSchedule`. `cronJobDefinition.setup` then reused that cached value. This meant:

- Every job built from the same `JobDefinition` (via `s.NewJob(def, ...)` calls) shared a single `*defaultCron` pointer.
- Calling `job.Update(...)` on one of those jobs re-invoked `IsValid` on the shared receiver, racing with any concurrent `Job.NextRuns()` call on a *different* job that also derived from the same definition.

**After:** `IsValid` is now a pure validator. `cronJobDefinition.setup` calls `IsValid` for its error side effect and separately owns a freshly-parsed `cron.Schedule` on the newly-created `cronJob`. Each `cronJob` holds its own parsed schedule; definitions can safely be reused across `NewJob` calls.

## H5 — `durationRandomJob` uses `math/rand/v2`

**Before:** `durationRandomJob` held a `*rand.Rand` (from `math/rand`, not `math/rand/v2`). The stdlib `rand.Rand` is not safe for concurrent use. The scheduler goroutine and user callers of `Job.NextRuns()` both invoke `next()`, which reads from the shared `*rand.Rand` — a race that `-race` reliably tripped once you hit it hard enough.

**After:** Switched to `math/rand/v2` and dropped the explicit `*rand.Rand` field. The v2 top-level functions use a per-goroutine PRNG that is safe for concurrent use by design.

## Regression tests

- `TestScheduler_CronJob_DefinitionReuseDoesNotAliasCronImpl` — asserts that two jobs built from the same `CronJob(...)` definition hold distinct `cron.Schedule` pointers. Uses a test-only concrete-type cast to reach into the scheduler's job map.
- `TestScheduler_DurationRandomJob_NextRunsIsRaceFree` — hammers `Job.NextRuns` from 8 goroutines while the scheduler is running. Reliably tripped `-race` before the fix; passes cleanly after.

## Verified

- `go build ./...`: clean
- `golangci-lint run` (with `modernize` + `embeddedstructfieldcheck`): 0 issues
- `go test -race -count=1 ./...`: pass
- Pre-commit hooks pass

## Context

This is Plan #4 from the code-review sweep — landed on branch `state-hygiene-c4-h5` (rebased onto current `v2` including #935 and #936). Fills the gap between #932 (C3) and #933 (H4) in the review numbering. See `CODE_REVIEW.md` C4 and H5 for the original findings.

## Non-goals

- **No API changes.** `Cron.IsValid` still returns `error`; the removal of the receiver mutation is an implementation detail invisible at the interface boundary. Custom `Cron` implementations that were coincidentally relying on `IsValid`-side-caching for `Next()` performance should switch to parsing once in their own constructor.

### DIFF
--- a/job.go
+++ b/job.go
@@ -296,7 +296,7 @@ func (c cronJobDefinition) setup(j *internalJob, location *time.Location, now ti
 		// by IsValid isn't shared across jobs derived from the same
 		// JobDefinition, and isn't concurrently mutated by later
 		// setups (e.g. Update) while another goroutine is reading
-		// through Job.NextRuns. See C4 in the code review.
+		// through Job.NextRuns.
 		cronImpl = &defaultCron{withSeconds: dc.withSeconds}
 	}
 
--- a/job.go
+++ b/job.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
+	"math/rand/v2"
 	"slices"
 	"strings"
 	"time"
@@ -180,6 +180,14 @@ type limitRunsTo struct {
 // Next returns the next scheduled run after lastRun. Callers assume the
 // returned time is strictly after lastRun; returning lastRun or an earlier
 // value can cause the scheduler to spin.
+//
+// If a custom implementation caches parsed state in IsValid for later
+// use by Next, it must either be safe for concurrent use across the
+// goroutines that call NewJob/Update, next-computation in the
+// scheduler, and Job.NextRuns from user code, or the caller must
+// supply a fresh instance per job via WithCronImplementation. The
+// default implementation is cloned per job automatically to avoid
+// aliasing when the same JobDefinition is reused across NewJob calls.
 type Cron interface {
 	IsValid(crontab string, location *time.Location, now time.Time) error
 	Next(lastRun time.Time) time.Time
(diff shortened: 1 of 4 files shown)


job.go
```diff
@@ -280,15 +288,23 @@ type cronJobDefinition struct {
...
+	cronImpl := c.cron
 	if j.cron != nil {
-		c.cron = j.cron
+		cronImpl = j.cron
+	} else if dc, ok := cronImpl.(*defaultCron); ok {
+		// Give each job its own defaultCron so parsing state written
+		// by IsValid isn't shared across jobs derived from the same
+		// JobDefinition, and isn't concurrently mutated by later
+		// setups (e.g. Update) while another goroutine is reading
+		// through Job.NextRuns. See C4 in the code review.
```
```suggestion
		// through Job.NextRuns.
```

scheduler_test.go
```diff
@@ -3468,3 +3467,70 @@ func TestScheduler_NextRuns_ReturnsAscendingAfterRescheduleCycles(t *testing.T)
...
 	require.NoError(t, s.Shutdown())
 }
+
+// TestScheduler_CronJob_DefinitionReuseDoesNotAliasCronImpl asserts
+// that reusing a JobDefinition across multiple NewJob calls yields
+// jobs with independent Cron implementations. Previously, all jobs
+// derived from the same definition shared the same *defaultCron
+// pointer, which was mutated by IsValid during setup — creating a
+// latent data race when Update ran concurrently with Job.NextRuns.
+// See C4 in CODE_REVIEW.md.
```
```suggestion
```

scheduler_test.go
```diff
@@ -3468,3 +3467,70 @@ func TestScheduler_NextRuns_ReturnsAscendingAfterRescheduleCycles(t *testing.T)
...
+		"each job derived from the same JobDefinition must hold its own Cron instance")
+
+	require.NoError(t, s.Shutdown())
+}
+
+// TestScheduler_DurationRandomJob_NextRunsIsRaceFree hammers
+// Job.NextRuns from many goroutines while the scheduler is running,
+// which previously tripped `-race` because durationRandomJob used a
+// non-concurrent *rand.Rand shared between the scheduler goroutine
+// and user callers of NextRuns. See H5 in CODE_REVIEW.md.
```
```suggestion
// and user callers of NextRuns.
```

